PR #1, opened 2016-02-20
Fix build errors, implement parseCommaList

sparkprime:
@jbeda


sparkprime:
My Go is pretty rusty, I probably already mentioned :)

PR #3, opened 2016-02-21
Fix up repo paths in the README.

jbeda:
@sparkprime You'll probably have to set up travis and the coverage tool as I'm not an admin on the repo.  We also need to enable issues on this repo so that users can submit golang specific issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 [![GoDoc Widget]][GoDoc] [![Travis Widget]][Travis] [![Coverage Status Widget]][Coverage Status]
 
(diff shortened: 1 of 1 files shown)


sparkprime:
Issues now enabled.

I think for travis in the last repo I just added a .travis.yml and it worked, so maybe merging this will be enough.

@davidzchen @mikedanese do you know how to set up coverage or CLA widgets?

sparkprime:
Ah no, to enable travis I also had to turn it on in https://travis-ci.org/profile/

sparkprime:
CLA hook should be set up now.  Thanks Mike!

PR #7, opened 2016-02-28
Finish up blocking out the parser.

jbeda:
There is the most minimal of tests and so this probably doesn't work well yet.  Adding tests is the next job.


sparkprime:
Halfway through so far, hopefully will finish off on the subway tomorrow.

parser.go
```diff
@@ -132,6 +155,572 @@ func (p *parser) parseCommaList(end tokenKind, elementKind string) (*token, astN
...
+				expr:    cond,
+			})
+		}
+		if maybeIf.kind == end {
+			return &specs, maybeIf, nil
+		}
+
+		if maybeIf.kind != tokenFor {
+			return nil, nil, makeStaticError(
+				fmt.Sprintf("Expected fo, if or %v after for clause, got: %v", end, maybeIf), maybeIf.loc)
```
sparkprime: s/fo,/for,

jbeda:
Fixed up typo.  Will submit once tests pass.

As for next steps, I was going to do some tests for the parser.  If you want to start on the next level up (desurgaring?) that'd be great.  

Getting the interface set should be easy too but I'm not sure how we can make it do anything short of having something actually running.

sparkprime:
BTW I renamed desugaring.cpp to desugarer.cpp on the C++ one because of lexer and parser  I think I should rename vm.cpp to interpreter.cpp so I'll just use those names right off the bat here.

PR #8, opened 2016-03-08
Finish writing unit tests for parser

jbeda:
Our code coverage is at ~83%.  It would be higher except that the generated code doesn't have tests.


sparkprime:
LGTM

PR #9, opened 2016-03-08
First end-to-end test (add numbers)

sparkprime:
So I coded this pretty fast, I'm still rusty on Go conventions and such.  Writing code from scratch is much harder than modifying / extending existing code.  I suspect I am capitalizing things wrongly :)

I added this main.go as the public API, that calls through the whole pipe even though most stages in that pipe aren't implemented.

Filling in the desugarer will gradually allow us to write more complex Jsonnet programs.  In the mean time we can run programs that do not use those more complex ASTs without the desugarer.  So completing the desugarer can happen concurrently with completing the interpreter & analyser.

The static analyser rejects bad code, which means as long as we run good programs, we don't need it.  Completing the analyser can therefore happen concurrently as well.  The other thing the static analyser does is compute free variables, which are used to compute the environment when building a closure (or thunk).  This is an optimization (we don't want to copy the whole environment into the closure, just the variables the closure actually needs).  So when we come to implement closures and thunks we'll need at least that behavior from the static analyser (but the checking ./ rejecting of code can wait).

I decided to roll state.h into the interpreter because without all that GC code there's really not much code there, and it's joined at the hip with the interpreter (nothing else uses it, and there's a really heavy dependency).

Rather than using a tagged union to represent values in the VM as the C++ code does, I'm using struct pointers that all extend the value interface.  I think that might be slower than C++ (more memory allocation) but I'm not sure what the performance model is in Go.  Even if it is slower, it may not actually matter.  Since Go doesn't have unions I'm not sure we have a choice anyway?

The interpreter is using the native Go stacks as we discussed before -- the code will be an order of magnitude simpler this way!

I left some empty files as placeholders for unit tests.  With not much real behavior implemented, there didn't seem to be much point.


sparkprime:
Rusty isn't the right word, I basically haven't done anything this complex in Go before :)

jbeda:
I just landed in London so won't be able to look for a bit. However you
might want to run go vet and go lint on it to make sure the coding
conventions are followed.

Lots of editor extensions will do this for you on every save.

Joe

jbeda:
Feedback from `golint`:

```
interpreter.go:34:6: exported type RuntimeError should have comment or be unexported
interpreter.go:161:1: receiver name should be a reflection of its identity; don't use generic names such as "me", "this", or "self"
interpreter.go:161:33: don't use underscores in Go names; method parameter ast_ should be ast
interpreter.go:191:9: if block ends with a return statement, so drop this else and outdent its block
interpreter.go:199:1: receiver name should be a reflection of its identity; don't use generic names such as "me", "this", or "self"
interpreter.go:199:25: method manifestJson should be manifestJSON
interpreter.go:200:2: method parameter v_ should be v
literalfield_set.go:15:44: exported func NewliteralFieldSet returns unexported type jsonnet.literalFieldSet, which can be annoying to use
main.go:19:1: comment on exported type Vm should be of the form "Vm ..." (with optional leading article)
main.go:20:6: type Vm should be VM
main.go:26:1: exported function MakeVm should have comment or be unexported
main.go:26:6: func MakeVm should be MakeVM
main.go:33:1: exported method Vm.ExtVar should have comment or be unexported
main.go:37:1: exported method Vm.ExtCode should have comment or be unexported
main.go:41:1: exported method Vm.EvaluateSnippet should have comment or be unexported
```

sparkprime:
Will do.  I'm going to be busy the next few days but then I'll do another pass on it.  Enjoy London, I lived there for 5 years during my MSc & PhD :)

jbeda:
I'm going to merge this and fix up the lint issues.  Other than that, it looks great.

sparkprime:
Yeah sorry I never got around to it, very busy here with perf, and currently at the Deis office working on Helm for a hackathon.  I'll get back on track soon though!

PR #10, opened 2016-03-14
Fix up lint errors.

jbeda:
Also:
* Made all types used in errors public.
* Made interpreter receiver a pointer.  This will reduce GC load for deep stacks.


sparkprime:
LGTM

PR #12, opened 2016-04-04
Pass through JSON

### LINKED ISSUE #11: Pass through JSON
I've set myself an arbitrary objective today to allow passing through JSON, i.e. add support for desugaring, checking, executing, manifesting basic objects and arrays containing other objects / arrays / primitives.

sparkprime:
So the rabbit hole was deeper than I thought... But I believe this is all that is needed for objects, arrays, and strings.  No tests.  No regrets!

The desugarer is fully laid out but has lots of holes (marked with TODOs).  There is no static analysis yet!  This will become very important when evaluating functions and thunks, because it is the static analysis phase that fills in the freeVariables for each astNode.  Without that, variable lookup will fail if the variable is captured from the environment (instead of being a direct parameter).

Simple objects are added to the interpreter, these are objects formed by { } and not with a comprehension and or by combining two objects with the + operator (a mixin).  I left space to add the code for these, but it wasn't necessary for passing through JSON.

The stack is very different to in the C++ version in that it is only used for variable bindings and reconstructing stack traces in errors.  It is not used to store partial execution.  Likewise the inner interpreter loop is very different and much more readable as it uses simple recursion.

Fixes #11


sparkprime:
About unit tests and coverage:  For the interpreter I think we're good with end-to-end tests.  For the desugarer, some unit tests are probably a good idea for development purposes.  I'm pretty confident we can get good coverage all the way through with just end-to-end tests (or if not, we can construct more end-to-end tests to get good coverage).  However for development, it's nice to have tests that validate particular areas of the code.  However, I haven't done this yet.

jbeda:
I've started looking at this but don't have a ton of bandwidth right now (moving, vacation). Let me know if you need me to hurry up.

sparkprime:
Nah there's no particular pressure on this.  I intend to integrate Jsonnet with helm but for a first release of Helm it's fine to use the C++ code, and then we can swap it out for the Go version later when security becomes a concern.  I am quite happy with the progress on go-jsonnet so far.

ast.go
```diff
@@ -386,7 +392,15 @@ type astObjectField struct {
 	expr2, expr3  astNode     // In scope of the object (can see self).
 }
 
-// TODO(jbeda): Add constructor helpers here
+// TODO(jbeda): Add the remaining constructor helpers here
+
+func astObjectFieldLocal(methodSugar bool, id *identifier, ids identifiers, trailingComma bool, body astNode) astObjectField {
```
jbeda: linter: This function is currently unused.

desugarer.go
```diff
@@ -16,15 +16,387 @@ limitations under the License.
...
+
+		case astObjectFieldExpr:
+		// Nothing to do.
+
+		case astObjectFieldStr:
+			// Just set the flag.
+			field.kind = astObjectFieldExpr
+
+		case astObjectLocal:
+			return fmt.Errorf("INTERNAL ERROR: Locals should be removed by now.")
```
jbeda: desugarer.go:182:22:warning: error strings should not end with punctuation (golint)

interpreter.go
```diff
@@ -88,43 +97,63 @@ func makeValueNull() *valueNull {
...
+	return t.content != nil
 }
 
 type valueArray struct {
-	Elements []thunk
+	elements []*thunk
 }
 
-func makeValueArray(elements []thunk) *valueArray {
+func makeValueArray(elements []*thunk) *valueArray {
```
jbeda: warning: makeValueArray is unused (deadcode)

jbeda:
My editor is using [gometalinter](https://github.com/alecthomas/gometalinter) so that might be useful for annoying to remember golang stuff.

`gometalinter --vendor --fast --disable=gocyclo ./...`

Looking at code more carefully now.  Hopefully I'll be done by end of day.  Sorry it took me so long to pick this up again.

desugarer.go
```diff
@@ -16,15 +16,387 @@ limitations under the License.
...
+			continue
+		}
+		binds = append(binds, astLocalBind{variable: *local.id, body: local.expr2})
+	}
+	for _, field := range *fields {
+		if field.kind == astObjectLocal {
+			continue
+		}
+		if len(binds) > 0 {
+			field.expr2 = &astLocal{astNodeBase{loc: *field.expr2.Loc()}, binds, field.expr2}
```
jbeda: Do we want to create astLocal once and share it across all of the fields?  Are there places where we mutate it later?

desugarer.go
```diff
@@ -16,15 +16,387 @@ limitations under the License.
...
+
+	return nil
+}
+
+func desugar(astPtr *astNode, objLevel int) error {
+	ast := *astPtr
+	// TODO(dcunnin): Remove all uses of unimplErr.
+	unimplErr := makeStaticError(fmt.Sprintf("Desugarer does not yet implement ast: %s", reflect.TypeOf(ast)), *ast.Loc())
+
+	var err error // Make all recursive calls of the form err = desugar, rather than some being err := desugar
```
jbeda: I think you can define this variable as part of the return -- 

```
func foo(...) (err error)
```

Then you can either just `return` or still `return err`.

desugarer.go
```diff
@@ -16,15 +16,387 @@ limitations under the License.
...
+		}
+
+	case *astFunction:
+		err = desugar(&ast.body, objLevel)
+		if err != nil {
+			return err
+		}
+
+	case *astImport:
+		// Nothing to do.
```
jbeda: No need to un-escape `ast.file` here?

interpreter.go
```diff
@@ -50,7 +59,7 @@ func (err RuntimeError) Error() string {
 
 // Values and state
 
-type bindingFrame map[*identifier]thunk
+type bindingFrame map[identifier]*thunk
 
 type value interface {
```
jbeda: This is a pretty generic name across what is becoming a large package.  We might want to break out the various stages into their own packages.  This would make some innards exported but it might be worth it.  Something to think about later, I guess.

interpreter.go
```diff
@@ -156,35 +272,216 @@ func makeCallStack(limit int) callStack {
...
+		return &valueArray{elements}, nil
+
 	case *astBinary:
 		// TODO(dcunnin): Assume it's + on numbers for now
-		leftVal, err := i.execute(ast.left)
+		leftVal, err := i.evaluate(ast.left)
 		if err != nil {
 			return nil, err
 		}
 		leftNum := leftVal.(*valueNumber).value
```
jbeda: This type assertion will panic if this thing isn't a `valueNumber`.  Might want to check it and throw an error instead.

jbeda:
LGTM -- I'm a noob here but it looks reasonable.

desugarer.go
sparkprime: I made an issue for this https://github.com/google/jsonnet/issues/172

sparkprime:
Thanks for review.  Took me over a week to actually find time to act on it :)  Do you use vim?  I just tried gometalinter on the commandline but had problems with errors like this:

```
static_error.go:20:2:error: could not import fmt (reading export data: /usr/lib/go/pkg/linux_amd64/fmt.a: go archive does not begin with __.SYMDEF or __.GOSYMDEF) (gotype)
```

and then tons of other errors that I suspect are a knock-on from that.

jbeda:
@sparkprime -- I'm using Atom for this project -- just switched for fun.  The go support seems pretty good. Sorry I can't help you with vim.

gometalinter isn't a must -- it is probably overkill.  Just thought I'd point it out since some of the stuff looks useful.

PR #15, opened 2017-07-28
Allow adding strings

sbarzowski:


sbarzowski:
This is my hello world change in this repository.

@lwander, @sparkprime

interpreter.go
```diff
@@ -360,6 +360,21 @@ func (i *interpreter) objectIndex(loc *LocationRange, obj value, f string, offse
...
+func (i *interpreter) evalAdd(leftVal, rightVal value) (value, error) {
+	switch leftVal := leftVal.(type) {
+	case *valueNumber:
+		left := leftVal.value
+		right := rightVal.(*valueNumber).value
+		return makeValueNumber(left + right), nil
+	case *valueString:
+		left := leftVal.value
+		right := rightVal.(*valueString).value
+		return makeValueString(left + right), nil
```
lwander: I think a `default` is preferable to completing the `switch` without hitting a `case`.

main_test.go
```diff
@@ -36,6 +36,7 @@ var mainTests = []mainTest{
 	{"simple_arith1", "3 + 3", "6", ""},
 	{"simple_arith2", "3 + 3 + 3", "9", ""},
 	{"simple_arith3", "(3 + 3) + (3 + 3)", "12", ""},
+	{"simple_arith_string", "\"aaa\" + \"bbb\"", "\"aaabbb\"", ""},
```
lwander: I'd probably add a silly empty string concat check as well

interpreter.go
```diff
@@ -360,6 +360,21 @@ func (i *interpreter) objectIndex(loc *LocationRange, obj value, f string, offse
...
+	}
+	// TODO(sbarzowski) More types and more graceful error handling
+	panic("unknown type")
```
lwander: I know you plan to flush this out in the future, but a simple `panic(fmt.Sprintf("unknown type: %v", leftVal))` will probably avoid some potential headscratching.

interpreter.go
sbarzowski: done

main_test.go
sbarzowski: done

interpreter.go
sbarzowski: done

main_test.go
sbarzowski: However, this is not the primary place for tests. We'll probably share the test suite with the C++ version with examples in separate files. We don't do that yet, because command line interface is not implemented yet.

sparkprime:
Thank you for implementing this important feature

PR #16, opened 2017-08-03
[WIP] Indexing, locals and functions

sbarzowski:


sbarzowski:
It's a work in progress, submitted for some early feedback.

interpreter.go
```diff
@@ -26,9 +26,12 @@ import (
 
 // Misc top-level stuff
 
+// TODO what is it?
+// External variable (or code) provided before execution
+// TODO how do they work?
```
sparkprime: if you run jsonnet --ext-str foo=bar then value is "bar" and isCode is false.

desugarer.go
```diff
@@ -199,6 +199,17 @@ func desugarFields(location LocationRange, fields *astObjectFields, objLevel int
 	return nil
 }
 
+func desugarArrayComp(astComp *astArrayComp, objLevel int) (err error) {
```
sparkprime: If you like, I think you can use a bool for the objLevel (i.e. abstract it to == 0 or > 0).

interpreter.go
```diff
@@ -151,28 +158,153 @@ type valueSimpleObject struct {
...
+		// 	elThunk.upValues = i.capture(el.FreeVariables())
+		// 	elements = append(elements, elThunk)
+		// }
+		// return &valueArray{elements}, nil
+
+		self := obj
+		offset := 0
+		field := found.fields[f]
+
+		return makeThunk("???", self, offset, field.body), nil
```
sparkprime: you're executing it now, not later

interpreter.go
```diff
@@ -146,33 +176,203 @@ type valueSimpleObjectField struct {
...
+		return nil, makeRuntimeError(loc, fmt.Sprintf("Field does not exist: %s", f))
+	}
+	switch found := found.(type) {
+	case *valueSimpleObject:
+		self := found
+		offset := foundAt
+		upValues := bindingFrame{}
+		field := found.fields[f]
+
+		return makeThunk("???", makeEnvironment(upValues, self, offset), field.body), nil
```
sparkprime: The evaluation of the body happens immediately, so no need to make a thunk, just recurse,

interpreter.go
```diff
@@ -146,33 +176,203 @@ type valueSimpleObjectField struct {
...
+	// has been inherited.  When evaluating a field from one of these super objects,
+	// we need to bind self to the concrete object (so self must point
+	// there) but uses of super should be resolved relative to the object whose
+	// field we are evaluating.  Thus, we keep a second field for that.  This is
+	// usually 0, unless we are evaluating a super object's field.
+	// TODO(sbarzowski) provide some examples
+	// TODO(sbarzowski) provide somewhere a complete explanation of the object model
+	// How deep in super we are. If we're executing
+	// TODO(sbarzowski) I was confused when I saw this name? Is it a standard term?
+	// If not maybe something along the lines of superDepth would be more obvious.
```
sparkprime: superDepth is a better name

interpreter.go
```diff
@@ -146,33 +176,203 @@ type valueSimpleObjectField struct {
...
+	t.env = environment{}
+}
+
+func (t *thunk) filled() bool {
+	return t.content != nil
+}
+
+func (t *thunk) getValue(i *interpreter, loc *LocationRange) (value, error) {
+	if t.filled() {
+		// TODO(sbarzowski) what for? Stack trace?
```
sparkprime: I'm not sure.

interpreter.go
```diff
@@ -146,33 +176,203 @@ type valueSimpleObjectField struct {
...
-	bindings bindingFrame
+	// True if it switches to a clean environment (function call or thunk)
+	// False otherwise, e.g. for local
+	// This makes callFrame a misnomer as it is technically not always a call...
+	isCall bool
+
+	// The code we were executing before.
+	// TODO(sbarzowski) what is it? Before? I guess it's the place from where it's
+	// called etc.
+	// TODO(sbarzowski) how could it be nil?
```
sparkprime: I think there are some edge cases, like calling a Jsonnet function from a builtin (map / filter / makeArray).  Also perhaps manifestation?

interpreter.go
```diff
@@ -146,33 +176,203 @@ type valueSimpleObjectField struct {
...
+	ast astNode
+
+	// The location of the code we were executing before.
```
sparkprime: Even in the cases where the ast was null, I think there was still a useful LocationRange.

interpreter.go
```diff
@@ -271,15 +477,17 @@ func makeCallStack(limit int) callStack {
...
+// Keeps current execution context and evaluates things
 type interpreter struct {
-	stack          callStack
-	idArrayElement identifier
-	idInvariant    identifier
-	externalVars   vmExtMap
+	stack          callStack  // TODO what is it?
+	idArrayElement identifier // TODO what is it?
+	idInvariant    identifier // TODO what is it?
+	externalVars   vmExtMap   // TODO what is it?
```
sparkprime: The data provided by the user that is made available through std.extVar()

interpreter.go
```diff
@@ -271,15 +477,17 @@ func makeCallStack(limit int) callStack {
...
 // TODO(dcunnin): Add multi output.
 
+// Keeps current execution context and evaluates things
 type interpreter struct {
-	stack          callStack
-	idArrayElement identifier
-	idInvariant    identifier
-	externalVars   vmExtMap
+	stack          callStack  // TODO what is it?
+	idArrayElement identifier // TODO what is it?
```
sparkprime: Array elements are lazy, so I re-used the thunk logic for them.  However the thunk requires a variable name (it is used in error messages).  So these are placeholders to put in that slot when using the thunk for other purposes:  1) array elements 2) object invariants (asserts).  The text of the identifier is "array_element" and "object_assert" and appears in the following stacktraces:

```
$ jsonnet -e '[ error "foo" ]'
RUNTIME ERROR: foo
        <cmdline>:1:3-13        thunk <array_element>
        During manifestation
```

```
$ jsonnet -e '{ assert false }'
RUNTIME ERROR: Object assertion failed.
        <cmdline>:1:10-14       thunk <object_assert>
        During manifestation
```

sparkprime: Usually the text would have the variable, as so:

```
$ jsonnet -e 'local badger = error "foo"; badger'
RUNTIME ERROR: foo
        <cmdline>:1:16-26       thunk <badger>
        <cmdline>:1:29-34
```

sbarzowski:
Replaced by #24

PR #18, opened 2017-08-07
Basic commandline (so we can start testing compatibility)

sbarzowski:


sbarzowski:
This is a very basic version, just so that we can start running the tests in the most primitive way possible. Some questions for the future:
a) Should we use a library like https://github.com/spf13/cobra to implement full featured CLI or do it "manually" like in C++ version. If we want a library then which one?
b) How do we feel about having eval and fmt modes in one utility? Was it easier this way or is there a good reason why this is better?
c) How to run test_suite from C++ version on jenkins and dev machiens. In this change there is a file that is supposed to contain the path to C++ version. But maybe we should clone the whole google/jsonnet every time we check it on jenkins? Should we add a subrepo? What's the best/cleanest/most convenient way?

sbarzowski:
This change requires https://github.com/google/jsonnet/pull/336.

@sparkprime @lwander

jsonnet/cmd.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/google/go-jsonnet"
+)
+
+func usage() {
+	fmt.Println("usage: jsonnet <filename>")
+}
+
+func main() {
+	// TODO(sbarzowski) Be consistent about error codes with C++ maybe
```
lwander: I've always found it easiest to keep track of TODOS by creating an issue and putting the issue # into the TODO parens. Not something you have to adopt, but might be helpful.

jsonnet/cmd.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2016 Google Inc. All rights reserved.
```
lwander: Should probably be 2017

sparkprime:
@jbeda @hausdorff for opinions on go libraries and such

Pulling in the C++ repo as a submodule just for the tests sounds like a good idea.

jsonnet/cmd.go
sbarzowski: done

jsonnet/cmd.go
sbarzowski: Once the code stabilizes a bit this will be a very good option (at least for issues observable by the user).

PR #19, opened 2017-08-07
Fix bug messing up newlines

sbarzowski:
When the lexer backed up, no matter from what character it went
to the previous line.

### DIFF
--- a/lexer.go
+++ b/lexer.go
@@ -230,20 +230,18 @@ func checkWhitespace(a, b string) int {
 // ---------------------------------------------------------------------------
 // Lexer
 
+type position struct {
+	byteNo    int // Byte position of last rune read
+	lineNo    int // Line number
+	lineStart int // Rune position of the last newline
+}
+
 type lexer struct {
 	fileName string // The file name being lexed, only used for errors
 	input    string // The input string
 
-	pos        int // Current byte position in input
-	lineNumber int // Current line number for pos
-	lineStart  int // Byte position of start of line
-
-	// Data about the state position of the lexer before previous call to
-	// 'next'. If this state is lost then prevPos is set to lexEOF and panic
-	// ensues.
-	prevPos        int // Byte position of last rune read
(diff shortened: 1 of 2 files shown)


sbarzowski:
ping @lwander, @sparkprime

sparkprime:
I can't see what you changed here except refactoring to pull the 3 vars into a struct.  Somewhere in there, I'm sure there's a logic change too but I can't find it

lwander:
Backup used to restore the prior line number, now it just restores the prior "pos" struct: https://github.com/google/go-jsonnet/pull/19/files#diff-afa30dddc22eee32cb8f633eef30e7f2L305

sparkprime:
It used to restore the lineNumber, lineStart and byte pos, and now it does all 3 of those via struct assignment.

sparkprime:
Is this the change: https://github.com/google/go-jsonnet/pull/19/files#diff-afa30dddc22eee32cb8f633eef30e7f2R269

lwander:
You pointed out a logic change, but I don't think that was the intent.

Notice how before prevLineNumber was only being set on advancing over newlines: https://github.com/google/go-jsonnet/pull/19/files#diff-afa30dddc22eee32cb8f633eef30e7f2L279

In conjunction with what I pointed out before, it would always restore lineNumber to the last line. 

Another description of the bug would be "the prior token's line number was only incremented on newlines, but restored everytime a token was backed up"

sbarzowski:
Example string "a\nb".
And imagine the lexing goes like this (these are not real rules, just for seeing what next and backup does):
<op> <char> <line> <byteNo>
next "a" 1, 2
next "\n" 2, 2,
next "b" 2, 3
back "b" 1, 2 <--- do you see the problem?
next "b" 1, 3

sparkprime:
Yeah so backup() wasn't working before, but only because next() wasn't maintaining the counters correctly.

sparkprime:
LGTM

PR #20, opened 2017-08-08
Make it possible to evaluate stdlib

sbarzowski:
New features in parser and desugarer:
* Support for slices
* Support for methods
* Support for function sugar in locals
* Prevent crashes with obj/arr comprehensions

After this change evaluating stdlib doesn't crash.

### DIFF
--- a/desugarer.go
+++ b/desugarer.go
@@ -83,7 +83,8 @@ func stringUnescape(loc *LocationRange, s string) (string, error) {
 func desugarFields(location LocationRange, fields *astObjectFields, objLevel int) error {
 
 	// Desugar children
-	for _, field := range *fields {
+	for i := range *fields {
+		field := &((*fields)[i])
 		if field.expr1 != nil {
 			err := desugar(&field.expr1, objLevel)
 			if err != nil {
@@ -222,12 +223,26 @@ func desugarObjectComp(astComp *astObjectComp, objLevel int) (astNode, error) {
 	// TODO(sbarzowski) this
 }
 
+func buildSimpleIndex(obj astNode, member identifier) astNode {
+	return &astIndex{
+		target: obj,
+		id:     &member,
+	}
+}
+
+func buildStdCall(builtinName identifier, args ...astNode) astNode {
+	std := &astVar{id: "std"}
+	builtin := buildSimpleIndex(std, builtinName)
+	return &astApply{
+		target:    builtin,
+		arguments: args,
+	}
+}
+
 func desugar(astPtr *astNode, objLevel int) (err error) {
 	ast := *astPtr
 
 	if ast == nil {
-		// Nothing to desugar. This is normal when desugaring optional subtrees
-		// that happen to be empty.
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


sbarzowski:
@sparkprime @lwander

parser.go
```diff
@@ -944,18 +948,59 @@ func (p *parser) parse(prec precedence) (astNode, error) {
...
+						end = p.pop()
+						readyForNextIndex = true
+					} else if readyForNextIndex {
+						indexes[which], err = p.parse(maxPrecedence)
+						if err != nil {
+							return nil, err
+						}
+						readyForNextIndex = false
+					} else {
+						_, err := p.popExpect(tokenBracketR)
```
lwander: I'm missing something here, shouldn't the fact that the first `if p.peek().kind == tokenBracketR` failed mean that this `p.popExpect(tokenBracketR)` will always fail?

parser_test.go
```diff
@@ -200,6 +202,8 @@ var errorTests = []testError{
 
 	{`a[(b c)]`, `test:1:6-7 Expected token ")" but got (IDENTIFIER, "c")`},
 	{`a[b c]`, `test:1:5-6 Expected token "]" but got (IDENTIFIER, "c")`},
+	{`a[]`, `test:1:3-4 Index requires an expression`},
+	{`a[42:42:42:42]`, `test:1:11-12 Invalid slice: too many colons`},
```
lwander: Some non-error tests would be helpful too

desugarer.go
```diff
@@ -128,16 +128,21 @@ func desugarFields(location LocationRange, fields *astObjectFields, objLevel int
 
 	// Remove methods
 	// TODO(dcunnin): this
-	for _, field := range *fields {
+	for i := range *fields {
+		field := &((*fields)[i])
```
lwander: Why take the address?

desugarer.go
```diff
@@ -301,8 +335,33 @@ func desugar(astPtr *astNode, objLevel int) (err error) {
 	case *astIndex:
 		return unimplErr
 
+	case *astSlice:
+		stdslice := &astVar{id: "std.slice"}
+		var apply astNode
```
sbarzowski: I need *astNode (a pointer to astNode) and this is a different type from **astVar. And I didn't know I can do `var x type = y` :-). 

done

parser.go
sbarzowski: Yeah. I'm just using this function to get the right error message. I have changed this to be more explicit.

parser_test.go
sbarzowski: You are right, I forgot. And I have found (and fixed) a bug - missing :: handling. Thank you!

desugarer.go
sbarzowski: B/c it's modified in place. I don't want to make a copy, b/c then the changes won't be made in the "real" ast.

lwander: Ah I missed that, sounds good

parser.go
```diff
@@ -944,18 +960,58 @@ func (p *parser) parse(prec precedence) (astNode, error) {
 			op := p.pop()
 			switch op.kind {
 			case tokenBracketL:
-				index, err := p.parse(maxPrecedence)
-				if err != nil {
-					return nil, err
+				// handle slice
+				var indexes [3]astNode
+				which := 0
```
sparkprime: Maybe call it colons_consumed

sparkprime:
BTW I noticed coverage is steadily decreasing.  Can you keep that in mind as we will have to root out untested code at some point anyway.

PR #21, opened 2017-08-10
Add verbatim string support

sbarzowski:


sbarzowski:
@lwander @sparkprime

parser.go
```diff
@@ -346,6 +346,8 @@ func (p *parser) parseObjectRemainder(tok *token) (astNode, *token, error) {
 					kind:        astStringBlock,
 					blockIndent: next.stringBlockIndent,
 				}
+			// TODO(sbarzowski) are verbatim string literals allowed here?
```
sparkprime: Yes they should be allowed everywhere other string literals are allowed.

sparkprime: I see there is no test for it and it doesn't work in C++:

```
$ jsonnet -e '{@"foo": 3}'
STATIC ERROR: <cmdline>:1:2-7: Unexpected: VERBATIM_STRING_DOUBLE while parsing field definition
```

sparkprime:
LGTM

parser.go
sparkprime: It does work in import, but there is also no test for that

sparkprime: See also: https://github.com/google/jsonnet/issues/172

PR #23, opened 2017-08-15
Fix unicode escape sequences & some desugaring

sbarzowski:
The most important thing here is fixing a bug in parsing "\uABCD" and adding support for '\''.  

Other things:
Add comment from C++ version to desugarer
Provide mock desugaring for asserts, so they don't block on desugaring
Actual desugaring for super index
Change critical errors/bugs to panic instead of unimplError (no need to bother the code around with handling these - stack trace is better in such case, at the lowest layer we might recover from them and show a gentle message alongside the stack trace asking the user to report a bug here).


sbarzowski:
@sparkprime @lwander

sbarzowski:
> any reason why you put the unicode fix & the desugaring into one PR?
No "real" reason. These were just desugarer (string unescaping happens during desugaring) changes I made so that some more tests get to evaluation.

PR #24, opened 2017-08-16
Interpreter & runtime - minimal usable version

sbarzowski:
Accomplishments in this commit:
* Majority of language features implemented:
	* Unary operators
	* Binary operators
	* Conditionals
	* Errors
	* Indexing arrays
	* Indexing objects
	* Object inheritance
	* Imports
	* Functions
	* Function calls
* There is a quite nice way for creating builtins
* Static analyzer is there with most of the functionality
* Standard library is included and parts unaffected by missing features
work
* Some bugs in existing parts fixed
* Most positive tests from C++ version pass, the rest is failing mostly
due to missing builtins and comprehensions.
* Some initial structure was created that should allow more incremental
  and focused changes in the future PRs.
* Some comments/explanations added
* Panics translated to a little bit more gentle internal errors (with a
  link to issues on github).

What still sucks:
* Stack traces & error messages (there's some stuff in place)
* Almost everything is in the same package
* Stuff is exported or unexporeted randomly (see above)
* Missing a few lexing/parsing features
* Missing builtins
* Missing support for extvars and top-level-args
* Checking function arguments is missing
* No clean Go API that commandline and compatibility layer to C can use
* No compatibility layer to C
* Assertions don't work (desugaring level, assertEquals works).
* Manifestation stack traces (and generally it could use some work).
* The way environments are constructed is sometimes suboptimal/clumsy.


sbarzowski:
@sparkprime @lwander - here is the massive change I was talking about.

While it has some parts directly taken from C++ version, it is quite a bit different, so I think it's best to just read the whole "backend" (especially since it's most of it is new/changed).

Recommended reading order:
* Commit message
* value.go
* thunks.go
* interpreter.go
* builtins.go
* evaluator.go
* (everything else)

While it's in review I have some other stuff like finishing the parser and ast dumps, desugaring of assertions and exposing C API from go and some non-intrusive changes to this change (in particular adding more builtins).

I can't wait to see what you think.

builtins.go
```diff
@@ -0,0 +1,342 @@
+/*
+Copyright 2016 Google Inc. All rights reserved.
```
sbarzowski: done

runtime_error.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2016 Google Inc. All rights reserved.
```
sbarzowski: done

value.go
```diff
@@ -0,0 +1,377 @@
...
+// Various operations on values are allowed, depending on their type.
+// All values are of course immutable.
+type value interface {
+	aValue()
+
+	// TODO(sbarzowski) consider representing each type as golang object
+	typename() string
+}
+
+// potentialValue is something that may get evaluated to a value
```
sbarzowski: done, thx

builtins.go
```diff
@@ -0,0 +1,342 @@
...
+
+var uopBuiltins = []*UnaryBuiltin{
+	uopNot:        &UnaryBuiltin{name: "operator!", function: builtinNegation, parameters: identifiers{"x"}},
+	uopBitwiseNot: &UnaryBuiltin{name: "operator~", function: builtinBitNeg, parameters: identifiers{"x"}},
+	uopPlus:       &UnaryBuiltin{name: "operator+ (unary)", function: builtinIdentity, parameters: identifiers{"x"}},
+	uopMinus:      &UnaryBuiltin{name: "operator- (unary)", function: builtinUnaryMinus, parameters: identifiers{"x"}},
+}
+
+var funcBuiltins = map[string]evalCallable{
+	"length":          &UnaryBuiltin{name: "length", function: builtinLength, parameters: identifiers{"x"}},
```
lwander: Perhaps move the function names into constants?

error_formatter.go
```diff
@@ -0,0 +1,91 @@
...
+		"Please report a bug here: " + bugURL + "\n"
+}
+
+func (ef *ErrorFormatter) buildStackTrace(frames []TraceFrame) string {
+	// std::stringstream ss;
+	// ss << "RUNTIME ERROR: " << e.msg << std::endl;
+	// const long max_above = vm->maxTrace / 2;
+	// const long max_below = vm->maxTrace - max_above;
+	// const long sz = e.stackTrace.size();
+	// for (long i = 0; i < sz; ++i) {
```
lwander: This is a strange thing to leave as a comment, maybe instead link to this code in the google/jsonnet repo at a particular commit?

interpreter.go
```diff
@@ -267,150 +186,120 @@ func makeCallStack(limit int) callStack {
...
 		}
-		return &valueArray{elements}, nil
+		return makeValueArray(elements), nil
 
 	case *astBinary:
-		// TODO(dcunnin): Assume it's + for now
-		leftVal, err := i.evaluate(ast.left)
+		// Some binary operators are lazy, so thunks are needed in general
+		env := i.getCurrentEnv(ast)
+		left := makeThunk("x", env, ast.left)
```
lwander: "x" & "y" are thunk names? what are the names for? these don't seem very descriptive/helpful

error_formatter.go
sparkprime: https://github.com/google/jsonnet/blob/master/core/libjsonnet.cpp#L594

error_formatter.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type ErrorFormatter struct {
+	// TODO(sbarzowski) use this
+	// MaxStackTraceSize  is the maximum length of stack trace before cropping
+	MaxStackTraceSize int
+	// TODO(sbarzowski) use these
+	pretty   bool
+	colorful bool
+	SP       SourceProvider
```
sparkprime: let's strip these out if here and come up with a richer libjsonnet.so API, and then build it into both commandline clients (or move to a single commandline client or whatever)

error_formatter.go
```diff
@@ -0,0 +1,91 @@
...
+	var buf bytes.Buffer
+	for _, f := range frames {
+		fmt.Fprintf(&buf, "\t%v\t%v\n", &f.Loc, f.Name)
+		// TODO(sbarzowski) handle max stack trace size
+		// TODO(sbarzowski) I think the order of frames is reversed
+	}
+	return buf.String()
+}
+
+type SourceProvider interface {
```
sparkprime: remove it if not used

imports.go
```diff
@@ -0,0 +1,156 @@
...
+
+type Importer interface {
+	Import(codeDir string, importedPath string) (*ImportedData, error)
+}
+
+type ImportCacheValue struct {
+	err  error
+	data *ImportedData
+	// nil if we haven't tried to execute it yet
+	asCode potentialValue
```
sparkprime: There doesn't seem to be a need for asCode to be a potentialValue, it can be a value. If it's an error, you could put that in err.

imports.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type Importer interface {
+	Import(codeDir string, importedPath string) (*ImportedData, error)
+}
+
+type ImportCacheValue struct {
+	err  error
+	data *ImportedData
+	// nil if we haven't tried to execute it yet
```
sparkprime: // nil if we have only imported it via importstr?

interpreter.go
```diff
@@ -20,159 +20,70 @@ import (
...
+// for example it can be shared between array elements and function arguments
+type environment struct {
+	sb selfBinding
+
+	// Bindings introduced in this frame. The way previous bindings are treated
+	// depends on the type of a frame.
+	// If isCall == true then previous bindings are ignored (it's a clean
+	// environment with just the variables we have here).
+	// If isCall == false then if this frame doesn't contain a binding
+	// previous bindings will be used.
```
sparkprime: The bindings are actually composed, with most nested one winning, right?

interpreter.go
```diff
@@ -20,159 +20,70 @@ import (
...
-type valueSimpleObjectFieldMap map[string]valueSimpleObjectField
-
-type valueSimpleObject struct {
-	upValues bindingFrame
-	fields   valueSimpleObjectFieldMap
-	asserts  []astNode
-}
+type callFrame struct {
+	// True if it switches to a clean environment (function call or thunk)
+	// False otherwise, e.g. for local
```
sparkprime: the association between locals and thunks makes this a bit unclear, use another ast for hte example, like the conditional

interpreter.go
```diff
@@ -181,74 +92,82 @@ type callStack struct {
...
 func (s *callStack) pop() {
 	if s.top().isCall {
 		s.calls--
 	}
 	s.stack = s.stack[:len(s.stack)-1]
-
 }
 
+// TODO(sbarzowski) I don't get this. When we have a tail call why can't we just
+// pop the last call from stack before pushing our new thing.
```
sparkprime: I can't remember.  Feel free to try other approaches and make sure there is good test coverage.

Potential cases:

1) mark a call tailstrict when it's not a tail call
2) mark a call tailstrict when it's not being called from itself
3) mark a call tailstrict when the arguments are not actually strict

interpreter.go
```diff
@@ -422,19 +311,61 @@ func (i *interpreter) evaluate(a astNode) (value, error) {
 				// Omitted field.
 				continue
 			default:
-				return nil, makeRuntimeError(ast.Loc(), "Field name was not a string.")
+				return nil, e.Error("Field name was not a string.")
 			}
 
 			if err != nil {
```
sparkprime: this is already checked, above

interpreter.go
```diff
@@ -422,19 +311,61 @@ func (i *interpreter) evaluate(a astNode) (value, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		index, err := e.evalInCurrentContext(ast.index)
+		if err != nil {
+			return nil, err
+		}
+		switch target := targetValue.(type) {
+		case valueObject:
+			indexString := index.(*valueString).value
```
sparkprime: add todo for better error if type was not a string

interpreter.go
```diff
@@ -422,19 +311,61 @@ func (i *interpreter) evaluate(a astNode) (value, error) {
...
+			return target.index(e, indexString)
+		case *valueArray:
+			indexInt := int(index.(*valueNumber).value)
```
sparkprime: add todo for better error if type was not a number

interpreter.go
```diff
@@ -645,26 +612,112 @@ func (i *interpreter) manifestJSON(loc *LocationRange, v value, multiline bool,
...
+
+func buildStdObject(i *interpreter) (value, error) {
+	objVal, err := evaluateStd(i)
+	if err != nil {
+		return nil, err
+	}
+	obj := objVal.(*valueSimpleObject)
+	builtinFields := map[string]unboundField{
+		// TODO(sbarzowski) not a real thing, just the simples thing I could put there
+		"nullValue": &readyValue{&valueNull{}},
```
sparkprime: why?  looks like a left-over

interpreter.go
```diff
@@ -645,26 +612,112 @@ func (i *interpreter) manifestJSON(loc *LocationRange, v value, multiline bool,
...
+		importCache: MakeImportCache(importer),
+	}
+
+	stdObj, err := buildStdObject(&i)
+	if err != nil {
+		return nil, err
 	}
-	result, err := i.evaluate(ast)
+
+	i.initialEnv = makeEnvironment(
```
sparkprime: The empty selfBinding with poison value comes up twice.  Maybe refactor it to makeUnboundSelfBinding or even makeEmptyEnvironment (in which case you would assign the "std" field after it in this case)

runtime_error.go
```diff
@@ -0,0 +1,69 @@
...
+limitations under the License.
+*/
+
+package jsonnet
+
+// RuntimeError is an error discovered during evaluation of the program
+type RuntimeError struct {
+	StackTrace []TraceFrame
+	Msg        string
+	trace      *TraceElement
```
sparkprime: remove, not set anywhere

runtime_error.go
```diff
@@ -0,0 +1,69 @@
...
+		tf.Name = trace.context.Name
+	} else {
+		tf.Name = ""
+	}
+	return tf
+}
+
+type TraceContext struct {
+	// Location of creation/definition.
+	Loc *LocationRange
```
sparkprime: I don't think you need this.  It is possible (for a human at least) to derive it from the next line up in the stack trace.  This is true of the name too, but at least the name makes the stacktrace a lot easier to read, because you can eyeball the nesting of calls to confirm it makes sense.  You can't really eyeball LocationRange, you have to cross-reference the code.

error_formatter.go
sbarzowski: This is just additional context for the todos below. This convention was used when I joined the project and it was quite convenient.

I'm gonna replace it with a link provided by @sparkprime now.

error_formatter.go
sbarzowski: a) For now I think we won't expose it in libjsonnet api. We can add it in the future. For now it's just very convenient for me now and long-term I think it's also useful for users so I want it to start capitalizing early.
b) I already have a branch which starts to use this stuff.

error_formatter.go
sbarzowski: (as above)

interpreter.go
sbarzowski: Not sure what you mean by composed. The lookup starts at the most nested one and then we try outer ones. We finish when we reach a "call" frame.

thunks.go
```diff
@@ -0,0 +1,169 @@
...
+// -------------------------------------
+
+// thunk holds code and environment in which the code is supposed to be evaluated
+type thunk struct {
+	name identifier
+	env  environment
+	body astNode
+}
+
+func makeThunk(name identifier, env environment, body astNode) *cachedThunk {
```
sparkprime: makeThunk returning a cachedThunk is weird.  Maybe call thunk 'exprThunk' (or astThunk but then it looks like an AST node).  Then call cachedThunk just thunk?  Or, call this makeCachedExprThunk because that's what it really is.

thunks.go
```diff
@@ -0,0 +1,169 @@
...
+		argThunks[closure.function.parameters[i]] = arg
+	}
+
+	calledEnvironment := makeEnvironment(
+		addBindings(closure.env.upValues, argThunks),
+		closure.env.sb,
+	)
+	context := TraceContext{
+		Loc:  closure.function.Loc(),
+		Name: "function <anonymous>",
```
sparkprime: Add TODO: better function names and object names

interpreter.go
sbarzowski: You sure? Maybe I'm not seeing something, but I've added (not pushed yet) a test that actually gets there...

builtins.go
sbarzowski: todo added

error_formatter.go
sbarzowski: (done)

interpreter.go
sbarzowski: TODO added

interpreter.go
sbarzowski: done

interpreter.go
sbarzowski: referenced this comment in todo

interpreter.go
sbarzowski: done

interpreter.go
sbarzowski: (added one todo for both string and number)

interpreter.go
sbarzowski: right, removed

runtime_error.go
sbarzowski: done

runtime_error.go
sbarzowski: done

imports.go
sbarzowski: > it can be a value. If it's an error, you could put that in err.
A "problem" with that is that we would cache an error for importstr. This is only a theoretical issue now, because the error is fatal anyway so we won't get to importstr. But I think it will be cleaner when I move err to the data - b/c it's about being unable to even get the raw data.

interpreter.go
sbarzowski: Done.

BTW Currently we have 3 places in which we do a very similar thing - import, std evaluation and main evaluation. I may try to do something about it later. Not sure if it won't just complicate things...

thunks.go
sbarzowski: done

thunks.go
sbarzowski: I like makeCachedExprThunk the most. But I want to think about it a bit more and I have another idea that I may want to try. So I've added a TODO with your comment, I will look into it soon.

imports.go
sbarzowski: done

interpreter.go
sbarzowski: (test is already pushed `testdata/fieldname_not_string.input`)

thunks.go
sbarzowski: btw I've also changed the return type everywhere to potentialValue to express the intention better

interpreter.go
sparkprime: ok just update the comment to "doesn't contain a binding for a given variable" as I interpreted to mean the whole map is empty

PR #25, opened 2017-08-16
initialize the extMap

sevki:
Given the following scenario

	vm := jsonnet.MakeVM()
	vm.ExtVar("foo", "bar")

calling the `ExtVar` and `ExtCode` methods of a vm result in a panic
because the map has not been initialized, this patch adds
initializes the map.

Signed-off-by: Sevki <s@sevki.org>

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ func MakeVM() *VM {
 	return &VM{
 		MaxStack: 500,
 		MaxTrace: 20,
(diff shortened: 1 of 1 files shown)


sbarzowski:
LGTM @sparkprime

sparkprime:
Thanks for the contribution!

sevki:
thank you both for the quick turnaround :smile:

PR #30, opened 2017-08-25
More builtins

sbarzowski:


sbarzowski:
@sparkprime @lwander

sbarzowski:
hmm... coverage decreased, but this stuff is tested in "external" tests... hmmm... Maybe I should add individual single-line tests for each of these functions...

builtins.go
```diff
@@ -248,8 +249,79 @@ func builtinType(e *evaluator, xp potentialValue) (value, error) {
...
+	if math.IsNaN(x) {
+		return nil, e.Error("Not a number")
+	}
+	if math.IsInf(x, 0) {
+		return nil, e.Error("Overflow")
+	}
+	return makeValueNumber(x), nil
+}
+
+// TODO(sbarzowski) perhaps it is too magical for Go style
```
lwander: I like it

sparkprime: these kinds of abstractions are basically mandatory given the error boilerplate

PR #32, opened 2017-08-25
Parser & desugarer support for "in"

sbarzowski:


sbarzowski:
@lwander @sparkprime

parser/parser.go
```diff
@@ -1069,15 +1075,23 @@ func (p *parser) parse(prec precedence) (ast.Node, error) {
...
-					return nil, err
-				}
-				lhs = &ast.Binary{
-					NodeBase: ast.NewNodeBaseLoc(locFromTokenAST(begin, rhs)),
-					Left:     lhs,
-					Op:       bop,
-					Right:    rhs,
+				if op.kind == tokenIn && p.peek().kind == tokenSuper {
+					super := p.pop()
+					lhs = &ast.InSuper{
```
lwander: Why does this operations need special handling?

sbarzowski: super is not a _real_ object, but a selfBinding, using the nomenclature from the code.

Anyway it's a special construct in jsonnet syntax - super is not a standalone expression and instead there are 3 specific ways to use it:
* `super.id`
* `super[expr]`
* `expr in super`

PR #33, opened 2017-08-25
Optional args parser/lexer/ast support

sbarzowski:
Just "frontend" part. They are ignored during actual execution.


sbarzowski:
@lwander @sparkprime

parser/parser.go
```diff
@@ -162,6 +160,87 @@ func (p *parser) parseCommaList(end tokenKind, elementKind string) (*token, ast.
...
+			namedArgumentAdded = true
+			args.Named = append(args.Named, ast.NamedArgument{Name: *id, Arg: expr})
+		}
+		gotComma = false
+		first = false
+	}
+}
+
+// TODO(sbarzowski) - this returned bool is weird
+func (p *parser) parseParameters(end tokenKind, elementKind string) (*token, *ast.Parameters, bool, error) {
```
lwander: Why return *token when it seems to get ignored by each invocation?

parser/parser.go
```diff
@@ -162,6 +160,87 @@ func (p *parser) parseCommaList(end tokenKind, elementKind string) (*token, ast.
...
+	expr, err := p.parse(maxPrecedence)
+	if err != nil {
+		return nil, nil, err
+	}
+	return id, expr, nil
+}
+
+// TODO(sbarzowski) - this returned bool is weird
+// TODO(sbarzowski) - name - it's also used for parameters
+func (p *parser) parseArguments(end tokenKind, elementKind string) (*token, *ast.Arguments, bool, error) {
```
lwander: As far as I can tell `end` is always tokenParenR, why take this as a parameter?

parser/parser.go
```diff
@@ -162,6 +160,87 @@ func (p *parser) parseCommaList(end tokenKind, elementKind string) (*token, ast.
...
+	args := &ast.Arguments{}
+	gotComma := false
+	namedArgumentAdded := false
+	first := true
+	for {
```
lwander: You could break this loop out into two by first parsing only positional and then only named arguments. With a little cleverness you could get rid of the `first` variable as well. IMO it'll be a little more readable, since your cutting down from 3 booleans being toggled in a single loop down to 1.

parser/parser.go
lwander: Name could be "parseCommaList"

parser/parser.go
sbarzowski: parseParameters consumes parameters and the closing paren. This closing paren is potentially interesting for its location (it is used in parseArguments). 

For now I just removed it. 

Perhaps it shouldn't consume the closing paren (it can just peek). Opening paren is consumed by the caller, so perhaps it would be more expected. Or perhaps it should consume both? Maybe the location could be saved in ast.Arguments and we would avoid the weirdness of returning the token from parseArguments.

parser/parser.go
sbarzowski: right, done

sbarzowski: Hmm... it was like that before, but it's not just a comma-separated list of expressions. It parses a special format of arguments and parameters.

parser/parser.go
sbarzowski: I tried it and I was able to simplify it a bit, but I wasn't able to do what you propose nicely. Perhaps I'm missing something, but here are the problems:
* gotComma is needed as it, b/c we return it and save it in the AST. So however we could eliminate `gotComma` and `first` in favor of something like `readyForNextArgument` we would still need to reconstruct `gotComma`. 
* We don't know ahead of time if the next argument is going to be positional or not, so we would need to pass the first named argument to the second loop somehow and we would need to handle both cases in both loops anyway.

parser/parser.go
```diff
@@ -114,54 +114,99 @@ func (p *parser) peek() *token {
 	return &p.t[p.currT]
 }
 
-func (p *parser) parseIdentifierList(elementKind string) (ast.Identifiers, bool, error) {
-	_, exprs, gotComma, err := p.parseCommaList(tokenParenR, elementKind)
-	if err != nil {
-		return ast.Identifiers{}, false, err
+func (p *parser) doublePeek() *token {
+	return &p.t[p.currT+1]
```
sparkprime: can this ever overrun the array?

sparkprime: i suppose only if peek() == EOF

sbarzowski: Yes.

parser/parser.go
```diff
@@ -114,54 +114,99 @@ func (p *parser) peek() *token {
...
+			gotComma = true
+		} else {
+			gotComma = false
+		}
+
 		first = false
 	}
 }
 
+// TODO(sbarzowski) - this returned bool is weird
```
sparkprime: Necessary to preserve the optional final comma in the AST.

PR #34, opened 2017-08-30
String indexing, change of string representation

sbarzowski:
Also std.pow


sbarzowski:
@lwander @sparkprime 
I need to add some tests for it.

builtins.go
```diff
@@ -90,7 +90,7 @@ func builtinGreater(e *evaluator, xp, yp potentialValue) (value, error) {
 		if err != nil {
 			return nil, err
 		}
-		return makeValueBoolean(left.value > right.value), nil
+		return makeValueBoolean(stringLessThan(right, left)), nil
```
lwander: Weird that `<` is desugared to not `>=`, but the only string comparison is `<`. I'd imagine for the non-desugared builtin to mirror the string comparison (stringGreaterThan)

builtins.go
```diff
@@ -318,13 +318,25 @@ func builtinObjectHasEx(e *evaluator, objp potentialValue, fnamep potentialValue
...
+func builtinPow(e *evaluator, basep potentialValue, expp potentialValue) (value, error) {
+	base, err := e.evaluateNumber(basep)
+	if err != nil {
+		return nil, err
+	}
+	exp, err := e.evaluateNumber(basep)
+	if err != nil {
+		return nil, err
+	}
+	return makeDoubleCheck(e, math.Pow(base.value, exp.value))
```
lwander: Have you looked into go's math.pow vs the C++ equivalent's behavior? I'd imagine this is where the two jsonnet's implementations might start to exhibit weird differences.

value.go
```diff
@@ -61,11 +61,57 @@ func (v *valueBase) aValue() {}
...
+	}
+	for i := 0; i < length; i++ {
+		if a.value[i] != b.value[i] {
+			return a.value[i] < b.value[i]
+		}
+	}
+	return len(a.value) < len(b.value)
+}
+
+func (s *valueString) length() int {
```
lwander: Nice

builtins.go
sbarzowski: Interesting. I couldn't find any obvious differences. Both C++ and Go documentation have long lists of special cases. The lists are different, but most likely equivalent.

I was able to find the implementation of pow in Go (https://github.com/golang/go/blob/master/src/math/pow.go#L13). It references FreeBSD C library and IEEE 754-2008. So I hope there is some standard for what it should do.

I couldn't find any mainstream C++ implementation to compare what it actually does.

Do you have any suggestions for how I can check it more thoroughly?

builtins.go
lwander: Good point

builtins.go
lwander: Ha sorry no suggestions, just wondering if this has come up yet. Perhaps we should make a list of weirdness we might expect in the future/should test given the opportunity.

builtins.go
```diff
@@ -233,7 +233,7 @@ func primitiveEquals(e *evaluator, xp potentialValue, yp potentialValue) (value,
 		if err != nil {
 			return nil, err
 		}
-		return makeValueBoolean(left.value == right.value), nil
+		return makeValueBoolean(string(left.value) == string(right.value)), nil
```
sparkprime: I think this needs to compare the two arrays.  Firstly it avoids the intermediate representation.  Also can you confirm that string([]rune) is a loss-less conversion (e.g. no canonicalization).

sbarzowski: Yes, I believe it's a lossless conversion:

> In Go, a string is in effect a read-only slice of bytes. If you're at all uncertain about what a slice of bytes is or how it works, please read the previous blog post; we'll assume here that you have.
>
> It's important to state right up front that a string holds arbitrary bytes. It is not required to hold Unicode text, UTF-8 text, or any other predefined format. As far as the content of a string is concerned, it is exactly equivalent to a slice of bytes.

Source: https://blog.golang.org/strings

> Go's approach to normalization
> 
> As mentioned in the strings blog post, Go does not guarantee that characters in a string are
> normalized. However, the go.text packages can compensate. 

Source: https://blog.golang.org/normalization

sbarzowski: done

value.go
```diff
@@ -61,11 +61,53 @@ func (v *valueBase) aValue() {}
...
+
+func (s *valueString) index(e *evaluator, index int) (value, error) {
+	if 0 <= index && index < s.length() {
+		return makeValueString(string(s.value[index])), nil
+	}
+	return nil, e.Error(fmt.Sprintf("Index %d out of bounds, not within [0, %v)", index, s.length()))
+}
+
+func concatStrings(a, b *valueString) *valueString {
+	var result []rune
```
sparkprime: Is it possible to give a size hint?

sbarzowski: Yep, I'll do that.

sbarzowski: done

PR #35, opened 2017-08-31
Array comprehensions

sbarzowski:
Implemented using flatMap.

The representation in AST was changed - now it expresses
the semantic structure better (ifs are attached to fors and nesting
is explicit).

Ifs in comprehensions are not supported yet, but they should be easy to add.


sbarzowski:
@sparkprime @lwander

builtins.go
```diff
@@ -170,6 +170,29 @@ func builtinMakeArray(e *evaluator, szp potentialValue, funcp potentialValue) (v
...
+	arr, err := e.evaluateArray(arrp)
+	if err != nil {
+		return nil, err
+	}
+	fun, err := e.evaluateFunction(funcp)
+	if err != nil {
+		return nil, err
+	}
+	num := int(arr.length())
+	var elems []potentialValue
```
sparkprime: any way to hint the array length?

sbarzowski: yes: 
`func make([]T, len, cap) []T`
so I could do something like:
`potentialValue := make([]potentialValue, 0, num)`

However in the end the array may end up much shorter than the one that is an argument to flatMap, so the conservative option would be to have 0 (or constant) initial capacity.

sbarzowski: We could also do an equivalent of shrink_to_fit from C++, probably in makeValueArray - then additional capacity won't be able to inflate memory usage in the end.

Ideally, in the future we would have a set of benchmarks, providing some target for optimization.

sparkprime: You can also do two passes.  This is fine for now though.

sbarzowski: done

PR #36, opened 2017-09-05
Add toString builtin

sbarzowski:
+ Bugfix in string unparsing


sbarzowski:
@sparkprime @lwander

interpreter.go
```diff
@@ -460,7 +463,7 @@ func unparseString(v string) string {
 	for _, c := range v {
 		switch c {
 		case '"':
-			buf.WriteString("\n")
+			buf.WriteString("\\\"")
```
lwander: Huh, was the `\n` a prior bug or is something special going on? Your changed string makes sense.

value.go
```diff
@@ -61,11 +61,57 @@ func (v *valueBase) aValue() {}
...
+	}
+	for i := 0; i < length; i++ {
+		if a.value[i] != b.value[i] {
+			return a.value[i] < b.value[i]
+		}
+	}
+	return len(a.value) < len(b.value)
+}
+
+func (s *valueString) length() int {
```
lwander: Wasn't this added in a recent PR?

value.go
```diff
@@ -61,11 +61,57 @@ func (v *valueBase) aValue() {}
...
 type valueString struct {
 	valueBase
-	value string
+	// We use rune slices instead of strings for quick indexing
+	value []rune
+}
+
+func (s *valueString) index(e *evaluator, index int) (value, error) {
+	runeIndex := 0
+	for _, r := range s.value {
```
lwander: Why not check if `index` is in the correct bounds, and if so, just index directly into `s.value`?

value.go
lwander: Ah the PR wasn't merged yet :P sorry

value.go
sbarzowski: Ah, when I was changing representation, I forgot to change that. Thanks (this will be changed in the other PR).

sbarzowski: done (in https://github.com/google/go-jsonnet/pull/34)

interpreter.go
sbarzowski: I think it was just a simple bug. There were no tests for that or anything. I don't see any reason to do that.

value.go
sbarzowski: Actually it's turns out I don't need #34 here, so it's now based on master. Sorry for the confusion.

PR #37, opened 2017-09-05
Add more builtins

sbarzowski:


sbarzowski:
@lwander @sparkprime

builtins.go
```diff
@@ -131,6 +173,21 @@ func builtinAnd(e *evaluator, xp, yp potentialValue) (value, error) {
...
 
+func builtinOr(e *evaluator, xp, yp potentialValue) (value, error) {
+	x, err := e.evaluateBoolean(xp)
+	if err != nil {
+		return nil, err
+	}
+	if x.value {
+		return x, nil
+	}
+	y, err := e.evaluateBoolean(yp)
```
lwander: I realize we want to short-circuit, but does this mean we would supress any should-be-static checks of the LHS when the RHS is true?

sbarzowski: If lhs evaluates to true then we don't evaluate rhs at all. We even have a test for that `or3.input`.

Not sure what you mean by _should-be-static checks_. We statically check for undeclared variables etc. in both arguments.

lwander: Ah that just shows my lack of understanding of the internals; I had thought a fair amount of static checking was done alongside evaluation.

builtins.go
```diff
@@ -286,6 +342,30 @@ var builtinAtan = liftNumeric(math.Atan)
...
+		if err != nil {
+			return nil, err
+		}
+		y, err := e.evaluateNumber(yp)
+		if err != nil {
+			return nil, err
+		}
+		xInt := int64(x.value)
+		yInt := int64(y.value)
+		// TODO(sbarzowski) What if the result is inf, nan or otherwise not allowed?
```
sparkprime: I don't think it's possible.  You might lose precision though, e.g.

```
$ jsonnet -e '(1 << 62) | 1'
4611686018427387904
```

sbarzowski: Right. This comment dates back to when I thought float64 binary represenation is taken as is, instead of being converted to an integer.

I've added this example to the test suite.

builtins.go
sparkprime: The static checking is limited to 1) not using variables that are not defined 2) not using $, self, super lexically outside of an object.  That is applied to both sides of a || b, however true || error "yolo" will not raise as that's a dynamic error.

sparkprime:
LGTM, ping me when you've resolved conflicts

sbarzowski:
done

sbarzowski:
@sparkprime

PR #38, opened 2017-09-07
Even more builtins

sbarzowski:
Fix a bunch of bugs:
- Reverse meaning of boolean argument to objectFieldsEx and objectHasEx
- Slice desugaring using `std.slice` instead of `slice` as a field name.

Support + on string and something else.
Support + on arrays

assertEqual should now work properly


sbarzowski:
@lwander @sparkprime

PR #60, opened 2017-09-12
std.codepoint, std.char + a bunch of tests

sbarzowski:


sbarzowski:
@sparkprime @lwander

builtins.go
```diff
@@ -373,6 +374,33 @@ func builtinType(e *evaluator, xp potentialValue) (value, error) {
 	return makeValueString(x.typename()), nil
 }
 
+// TODO(sbarzowski) not entierly clear what this is for
+const codepointMax = 0x110000
+
+func builtinChar(e *evaluator, xp potentialValue) (value, error) {
+	x, err := e.evaluateNumber(xp)
```
lwander: Unicode allots 21bits for codepoints https://en.wikipedia.org/wiki/Code_point

sbarzowski: Proper comment added.

PR #62, opened 2017-09-15
Object invariants

sbarzowski:


sbarzowski:
@sparkprime @lwander
Don't be scared by +2000 :-). That's just test output. Important part of one test (a regression test for a performance bug) is manifesting an array with 2000 elements.

value.go
```diff
@@ -297,14 +301,41 @@ func makeUnboundSelfBinding() selfBinding {
...
 type valueObjectBase struct {
 	valueBase
+	assertionError error
 }
 
 func (*valueObjectBase) typename() string {
 	return "object"
 }
 
+func (obj *valueObjectBase) assertionsChecked() bool {
```
lwander: I'm a little confused here - can't an assertion be checked without producing an error?

lwander: Ah heh, just read your "errNoError". Makes sense...

sbarzowski: I wish I had proper algebraic types available. Coming up with special values like that feels dirty... 

I've added another comment to minimize confusion.

desugarer.go
```diff
@@ -110,10 +110,23 @@ func desugarFields(location ast.LocationRange, fields *ast.ObjectFields, objLeve
...
+		field.Expr3 = nil
+		onFailure := &ast.Error{Expr: msg}
+		assertion := &ast.Conditional{
+			Cond:        field.Expr2,
+			BranchTrue:  &ast.LiteralBoolean{Value: true}, // ignored anyway
+			BranchFalse: onFailure,
+		}
+		field.Expr2 = assertion
 		/*
 			AST *msg = field.expr3
```
sparkprime: delete C++ code in comment

sbarzowski: done

value.go
```diff
@@ -317,7 +350,43 @@ type valueSimpleObject struct {
...
+		}
+		return nil
+	default:
+		panic(fmt.Sprintf("Unknown object type %#v", obj))
+	}
+}
+
+func checkAssertions(e *evaluator, obj valueObject) error {
+	if !obj.assertionsChecked() {
+		// Assertions may refer to the object that will normally
```
sparkprime: I don't think I ever proved that this was sound.  It seems OK though :)

value.go
```diff
@@ -317,7 +350,43 @@ type valueSimpleObject struct {
...
+}
+
+func checkAssertionsHelper(e *evaluator, obj valueObject, curr valueObject, superDepth int) error {
+	switch curr := curr.(type) {
+	case *valueExtendedObject:
+		err := checkAssertionsHelper(e, obj, curr.right, superDepth)
+		if err != nil {
+			return err
+		}
+		err = checkAssertionsHelper(e, obj, curr.left, superDepth+curr.right.inheritanceSize())
```
sparkprime: spaces around +

sparkprime: also this

value.go
sbarzowski: I think it is.  Each object either passes the assertion or not. So the only thing that matters for the final result (i.e. error or no error) is the set of objects to check. It doesn't matter when we check them. We check them as soon as we can, b/c that is the most predictable, obvious thing to the user.

This would get more complicated if we added some sort of error handling.

value.go
sbarzowski: This is how gofmt formats it:
https://github.com/golang/go/issues/1206

PR #64, opened 2017-09-19
Add support for % operator

sbarzowski:


sbarzowski:
This turned out to be a trivial one line change, b/c it's implemented in std.jsonnet, so I just needed to enable desugaring.

Some tests are added as well.

@sparkprime @lwander

PR #77, opened 2017-09-25
ExtVar support

sbarzowski:
Only through internal API. Also no TLA support at the moment.


interpreter.go
```diff
@@ -189,15 +189,16 @@ func makeCallStack(limit int) callStack {
...
-// TODO(dcunnin): Add string output.
 // TODO(dcunnin): Add multi output.
 
 // Keeps current execution context and evaluates things
 type interpreter struct {
-	stack          callStack      // TODO what is it?
-	idArrayElement ast.Identifier // TODO what is it?
-	idInvariant    ast.Identifier // TODO what is it?
-	externalVars   vmExtMap       // TODO what is it?
+	// Current stack. It
```
lwander: Missing word(s) after `It`

sbarzowski: Thanks, fixed

imports.go
```diff
@@ -76,31 +76,34 @@ func (cache *ImportCache) ImportString(codeDir, importedPath string, e *evaluato
 	return makeValueString(data.data.content), nil
 }
 
+func codeToPV(e *evaluator, filename string, code string) potentialValue {
+	node, err := snippetToAST(filename, code)
+	if err != nil {
+		// TODO(sbarzowski) perhaps we should wrap (static) error here
```
sparkprime: It should be a runtime error.  Current behavior of C++ is broken.  We need to document and test
each case (import, importstr, ext-code, tla-code).

It should be allowed to import from a non-existent file in a branch that never gets executed.  That is the semantics of the language.

sparkprime: So yeah I think if you want to eagerly pre-parse the AST, you have to defer any syntax error until actual use.  The C++ implementation re-parses it every time it's referred to I think?

sbarzowski: > The C++ implementation re-parses it every time it's referred to I think?

No. I think Go and C++ implementation handle caching in pretty much the same way:

```cpp
/** Import another Jsonnet file.
     *
     * If the file has already been imported, then use that version.  This maintains
     * referential transparency in the case of writes to disk during execution.  The
     * cache holds a thunk in order to cache the resulting value of execution.
     *
     * \param loc Location of the import statement.
     * \param file Path to the filename.
     */
    HeapThunk *import(const LocationRange &loc, const LiteralString *file)
    {
        ImportCacheValue *input = importString(loc, file);
        if (input->thunk == nullptr) {
            // EVALUATION CODE
        }
        return input->thunk;
    }

/** Import a file as a string.
     *
     * If the file has already been imported, then use that version.  This maintains
     * referential transparency in the case of writes to disk during execution.
     *
     * \param loc Location of the import statement.
     * \param file Path to the filename.
     * \param found_here If non-null, used to store the actual path of the file
     */
    ImportCacheValue *importString(const LocationRange &loc, const LiteralString *file)
    {
        std::string dir = dir_name(loc.file);

        const UString &path = file->value;

        std::pair<std::string, UString> key(dir, path);
        ImportCacheValue *cached_value = cachedImports[key];
        if (cached_value != nullptr)
            return cached_value;
        // [code that actually fetches it]
        cachedImports[key] = input_ptr;
        return input_ptr;
    }
```

sbarzowski: Updated TODO to reflect the decision.

PR #79, opened 2017-09-25
Separate manifestation from serialization to string

sbarzowski:
This is necessary for example for native functions
(which take json as arguments).

Standard "encoding/json" representation is used, but I have
mixed feeling about it. Not sure if treating json values as interface{}
is the right trade-off in our case.

### DIFF
--- a/interpreter.go
+++ b/interpreter.go
@@ -692,10 +692,6 @@ func serializeJSON(v interface{}, multiline bool, indent string, buf *bytes.Buff
 	}
 }
 
-// TODO(sbarzowski) Perhaps it should be a builtin?
-// TODO(sbarzowski) Perhaps we should separate recursive evaluation from serialization?
-// 					Strictly evaluating something may be useful by itself.
-//					For example may help with error reporting from custom serialization functions.
 func (i *interpreter) manifestAndSerializeJSON(trace *TraceElement, v value, multiline bool, indent string) (string, error) {
 	var buf bytes.Buffer
 	manifested, err := i.manifestJSON(trace, v)
--- a/interpreter.go
+++ b/interpreter.go
@@ -525,15 +525,83 @@ func unparseNumber(v float64) string {
 	return fmt.Sprintf("%.17g", v)
 }
 
-// TODO(sbarzowski) Perhaps it should be a builtin?
-// TODO(sbarzowski) Perhaps we should separate recursive evaluation from serialization?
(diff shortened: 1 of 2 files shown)


sbarzowski:
@lwander @sparkprime

interpreter.go
```diff
@@ -503,15 +503,83 @@ func unparseNumber(v float64) string {
...
+	default:
+		return nil, makeRuntimeError(
+			fmt.Sprintf("Manifesting this value not implemented yet: %s", reflect.TypeOf(v)),
+			i.getCurrentStackTrace(trace),
+		)
+
+	}
+}
+
+func serializeJSON(v interface{}, multiline bool, indent string, buf *bytes.Buffer) {
```
lwander: Why can't we rely on `encoding/json` to serialize the interface here?

lwander:
What are your concerns with `interface{}`?

interpreter.go
sbarzowski: Because I didn't want to change the behavior of multiline/indentation. Having exactly the same output is very valuable for testing against C++ version. Unless "encoding/json" can be convinced to format it in exactly the same way, but I haven't look into that yet.

sbarzowski:
interface{} means that any type whatsoever can be used there. If I was doing it from scratch I would add some tag to make sure that only a fixed set of types can be used when json is expected. Also I don't think it's a big problem, just a bad gut feeling.

interpreter.go
lwander: That makes sense

interpreter.go
```diff
@@ -621,17 +662,26 @@ func (i *interpreter) manifestJSON(trace *TraceElement, v value, multiline bool,
...
-		return makeRuntimeError(
-			fmt.Sprintf("Manifesting this value not implemented yet: %s", reflect.TypeOf(v)),
-			i.getCurrentStackTrace(trace),
-		)
+		panic(fmt.Sprintf("Unsupported value for serialization %#+v", v))
+	}
+}
 
+// TODO(sbarzowski) Perhaps it should be a builtin?
+// TODO(sbarzowski) Perhaps we should separate recursive evaluation from serialization?
```
sparkprime: TODO can go now :)

sparkprime:
ping me for merge

sbarzowski:
@sparkprime ping

PR #81, opened 2017-09-25
Support for various kinds of strings in imports

sbarzowski:


sbarzowski:
@sparkprime @lwander

parser/parser.go
```diff
@@ -715,7 +715,7 @@ func (p *parser) parseTerminal() (ast.Node, error) {
 		return &ast.LiteralString{
 			NodeBase:    ast.NewNodeBaseLoc(tok.loc),
 			Value:       tok.data,
-			Kind:        ast.StringDouble,
```
sparkprime: Did this bug get missed by a test?

sbarzowski: This is pretty subtle, b/c I think we treat them exactly the same after parsing with one exception - imports. And there is no unparsing in Go version, so it didn't affect the behavior until now.

PR #87, opened 2017-09-27
Support of +:, in, in super

sbarzowski:
Also profiling


sbarzowski:
@sparkprime @lwander

It's a bit different from what happens in C++:
- For in and in super, C++ retrieves all fields and checks if any matches the right one. Here we just do a regular lookup for a field and then check if it's nil.
- +: is not desugared to a regular field. This way we avoid multiple evaluations of the key and object field lookups.

sbarzowski:
@sparkprime ping

PR #90, opened 2017-09-28
Fix comprehensions

sbarzowski:
Only first and last forspec was used due to a parser bug.


sbarzowski:
@sparkprime @lwander

parser/parser.go
```diff
@@ -537,21 +537,21 @@ func (p *parser) parseObjectRemainder(tok *token) (ast.Node, *token, error) {
 }
 
 /* parses for x in expr for y in expr if expr for z in expr ... */
-func (p *parser) parseComprehensionSpecs(end tokenKind) (*ast.ForSpec, *token, error) {
+func (p *parser) parseComprehensionSpecs(end tokenKind) (*ast.ForSpec, *ast.ForSpec, *token, error) {
```
lwander: The added return is only used inside this function from recursive calls, right? Does it make sense to expose to an external caller?

sbarzowski: You're right. I thought it was not worth it to, since it's an internal thing used in two places. 

I've changed it, let me know if you think it's better.

parser/parser.go
```diff
@@ -538,52 +538,57 @@ func (p *parser) parseObjectRemainder(tok *token) (ast.Node, *token, error) {
 
 /* parses for x in expr for y in expr if expr for z in expr ... */
```
sparkprime: The 'for' is already gone isn't it?

sbarzowski: Right.

parser/parser.go
```diff
@@ -538,52 +538,57 @@ func (p *parser) parseObjectRemainder(tok *token) (ast.Node, *token, error) {
...
 func (p *parser) parseComprehensionSpecs(end tokenKind) (*ast.ForSpec, *token, error) {
-	var ifSpecs []ast.IfSpec
+	var parseComprehensionSpecsHelper func() (*ast.ForSpec, *ast.ForSpec, *token, error)
+	parseComprehensionSpecsHelper = func() (*ast.ForSpec, *ast.ForSpec, *token, error) {
```
sparkprime: if you pass in 'outer' (can be nil) then you can drop the extra return param and do it in the ForSpec constructor, which is a bit simpler overall.

PR #94, opened 2017-09-29
eliminate unused code:

mqliang:
operator% is implemented in std.jsonnet and we already enable desugaring in https://github.com/google/go-jsonnet/pull/64. So those code can be safely removed.

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -655,13 +655,6 @@ func (b *TernaryBuiltin) Parameters() ast.Identifiers {
 	return b.parameters
 }
 
-func todoFunc(e *evaluator, x, y potentialValue) (value, error) {
-	return nil, e.Error("not implemented yet")
-}
-
-// so that we don't get segfaults
-var todo = &BinaryBuiltin{function: todoFunc, parameters: ast.Identifiers{"x", "y"}}
-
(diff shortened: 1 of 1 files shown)


builtins.go
```diff
@@ -670,9 +663,8 @@ var desugaredBop = map[ast.BinaryOp]ast.Identifier{
 }
 
 var bopBuiltins = []*BinaryBuiltin{
-	ast.BopMult:    &BinaryBuiltin{name: "operator*", function: builtinMult, parameters: ast.Identifiers{"x", "y"}},
-	ast.BopDiv:     &BinaryBuiltin{name: "operator/", function: builtinDiv, parameters: ast.Identifiers{"x", "y"}},
-	ast.BopPercent: todo,
+	ast.BopMult: &BinaryBuiltin{name: "operator*", function: builtinMult, parameters: ast.Identifiers{"x", "y"}},
+	ast.BopDiv:  &BinaryBuiltin{name: "operator/", function: builtinDiv, parameters: ast.Identifiers{"x", "y"}},
```
sbarzowski: Could you add a comment that it's desugared, like below for bopManifestEqual?

sbarzowski:
@sparkprime

sparkprime:
Thanks!

PR #98, opened 2017-09-29
Don't escape block strings

sbarzowski:
Previously I thought they would be normally escaped, but
C++ test suite proved me wrong.


sbarzowski:
@lwander @sparkprime

desugarer.go
```diff
@@ -518,15 +518,17 @@ func desugar(astPtr *ast.Node, objLevel int) (err error) {
 		// Nothing to do.
 
 	case *ast.LiteralString:
-		if node.Kind != ast.VerbatimStringDouble && node.Kind != ast.VerbatimStringSingle {
+		if node.Kind != ast.VerbatimStringDouble &&
+			node.Kind != ast.VerbatimStringSingle &&
```
sparkprime: Weird formatting here, is this the output of gofmt?  What happens if you begin the line with a &&

sbarzowski: I'm afraid it is.

>What happens if you begin the line with a &&

It's not valid Go. Go has some funny semicolon inference rules which dictate how you can end a line.

To avoid this awkwardness I extracted it to a method of StringKind.

PR #99, opened 2017-09-29
Allow all kinds of string literals as object field names

sbarzowski:


sbarzowski:
@sparkprime @lwander

sbarzowski:
(Tests failure related to #98)

Issue #100, opened 2017-10-02
Pre-parsed stdlib


sbarzowski:
@mqliang wrote in #89:
> It would be very interesting if we can dump ast as go source files, then we can dump ast of stdlib as go source code, and embed it into vm, thus avoid parsing stdlib every time. Any idea?

I think we should look into it. It sound quite easy.

mqliang:
I tried to dump ast using go-spew (https://github.com/davecgh/go-spew), it's practicable: about 7w line source code is generated. 

The problem is: we need fork go-spew and modify its source code to make the generated files meet go's syntax.

sbarzowski:
My feeling is that the easiest and most maintainable way would be to do it by hand, i.e. have a function that does a type switch over ast nodes and formats each of them into a go source code representation. There are just ~30 AST node types IIRC. This will give us full control over the output and let us specially handle location information, aliased nodes and other complex cases. I think that with some helper functions the boilerplate could be minimal.

mqliang:
utter(github.com/kortschak/utter) works. 

The only problem is: we need make ALL fields of AST node as public, then the dumped source code can be put in any package. Otherwise, we must put the source code in the "ast" package.

mqliang:
https://github.com/google/go-jsonnet/issues/82 will also benefit from this.

PR #101, opened 2017-10-02
experiment dump ast to source code

mqliang:
utter(github.com/kortschak/utter) is powerful to dump go data structure as go source code.

To dump ast and make the source code usable in other package, we need make "loc" and "freeVariables" in NodeBase struct as public. Otherwise, the dumped source code can only be used in "ast" package.

This pr is just send out for further comment and feedback and demonstrates that pre-parse stdlib is practicable. If this way looks good, I will close this and send several elaborate PRs:
* make "loc" and "freeVariables" in NodeBase struct as public
* add a "Dump()"  function in "ast" package, which leveraging utter(github.com/kortschak/utter)
* add a dump command line tool which can read a jsonnet file and dump it's ast. The dump tool will call the "Dump()"  function in "ast" package
* using the dump tool to generate go source code for stdlib, and embed it into vm.

### DIFF
--- a/dump/README.md
+++ b/dump/README.md
@@ -1,0 +1,72 @@
+# dump
+
+Package dump can dump a Go data structure to Go source file, so that it can be statically embeded into other code.
+
+# Project Status
+* UnsafePointer is not supported yet
+* cyclic pointer is not supported yet
+
+# Implementation Notes
+
+## Aliase
+
+Primitive pointer and reused pointer will be aliased.
+
+### Primitive pointers
+```go
+var a = "hello world"
+
+var Obj = &struct {
+    Foo *string
+    Bar *string
+}{
+	Foo: &a,
+	Bar: &a,
+}
+
+dump.dump(Obj)
+
+```
+
+will output
+```go
+var p1Var = "hello world"
+var p1 = &p1Var
+var Obj = &struct { Foo *string; Bar *string }{
+	Foo: p1,
+	Bar: p1,
+}
+```
+
+### Reused pointers
+```go
+type Zeo struct {
+	A *string
+}
+
+var z = Zeo {
+	A: "hello world",
+}
+
+var Obj = &struct {
+	Foo *Zeo
+	Bar *Zeo
+}{
+	Foo: &z,
+	Bar: &z,
+}
+
+dump.dump(Obj)
+```
+
+will output
+
+```go
+var p1 = &Zeo{
+	A: "hello world",
+}
+var Obj = &struct { Foo *Zeo; Bar *Zeo }{
+	Foo: p1,
+	Bar: p1,
+}
+```
--- a/dump/README.md
+++ b/dump/README.md
@@ -4,13 +4,13 @@ Package dump can dump a Go data structure to Go source file, so that it can be s
 
 # Project Status
 * UnsafePointer is not supported yet
-* cyclic pointer is not supported yet
+* Cycles in the object graph are not supported yet
 
 # Implementation Notes
 
-## Aliase
+## Aliasing
 
-Primitive pointer and reused pointer will be aliased.
+Aliases to primitives and structs will be preserved in the objects described by the generated code.
 
 ### Primitive pointers
 ```go
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -1,0 +1,510 @@
+/*
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+
+package dump
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"reflect"
+	"regexp"
+	"sort"
+	"strconv"
+)
+
+var packageNameStripperRegexp = regexp.MustCompile("\\b[a-zA-Z_]+[a-zA-Z_0-9]+\\.")
+
+// Options represents configuration options for litter
+type Options struct {
+	StripPackageNames bool
+	HidePrivateFields bool
+	HomePackage       string
+	VariableName 	  string
+}
+
+// Config is the default config used when calling Dump
+var Config = Options{
+	StripPackageNames: false,
+	HidePrivateFields: true,
+	VariableName: "p0",
+}
+
+type dumpState struct {
+	w      io.Writer
+	depth  int
+	config *Options
+
+	allPointers     []uintptr
+	visitedPointers []uintptr
+
+	reusedPointers    []uintptr
+	primitivePointers []uintptr
+
+	currentPointerName string
+	homePackageRegexp  *regexp.Regexp
+}
+
+func (s *dumpState) indent() {
+	s.w.Write(bytes.Repeat([]byte("\t"), s.depth))
+}
+
+func (s *dumpState) newline() {
+	s.w.Write([]byte("\n"))
+}
+
+func (s *dumpState) dumpType(v reflect.Value) {
+	typeName := v.Type().String()
+	if s.config.StripPackageNames {
+		typeName = packageNameStripperRegexp.ReplaceAllLiteralString(typeName, "")
+	} else if s.homePackageRegexp != nil {
+		typeName = s.homePackageRegexp.ReplaceAllLiteralString(typeName, "")
+	}
+	s.w.Write([]byte(typeName))
+}
+
+func (s *dumpState) dumpSlice(v reflect.Value) {
+	s.dumpType(v)
+	numEntries := v.Len()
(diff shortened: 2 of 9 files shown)


mqliang:
ref https://github.com/google/go-jsonnet/issues/100

mqliang:
@sbarzowski It turns out that you are right: "the easiest and most maintainable way would be to do it by hand". utter does not handle:

* reused pointers. For, example 
```go
type foo struct {
		A int
	}

	var v0 = &foo{
		A: 1,
	}

	var bar = struct {
		p0 *foo
		p1 *foo
	}{
		p0 : v0,
		p1 : v0,
	}
```
will be mistakenly be dump by utter as:
```
var bar = struct {
		p0 : &foo{A:1},
		p1 : &foo{A:1},
	}
```
where p0 and p1 are two different pointer!

* primitive pointers: pointers pointing to primitive type. For example, a pointer pointing to a string "hello" should be dumped as:

```go
var p0Var = "hello"
var p0 = &p0Var
```

However, uttter dump is:
```go
var p0 = &"hello"
```
which NOT meet go syntax.

mqliang:
So, I hand write a dump function, which can well handle "reused pointer" and "primitive pointer"

mqliang:
cc @sparkprime

dump/utils.go
```diff
@@ -0,0 +1,97 @@
...
+package dump
+
+import (
+	"io"
+	"strconv"
+	"reflect"
+)
+
+func printBool(w io.Writer, value bool) {
+	if value {
```
sparkprime: why not `w.write([]byte(strconv.FormatBool(value)))`

dump/utils.go
```diff
@@ -0,0 +1,97 @@
...
+func printFloat(w io.Writer, val float64, precision int) {
+	w.Write([]byte(strconv.FormatFloat(val, 'g', -1, precision)))
+}
+
+func printComplex(w io.Writer, c complex128, floatPrecision int) {
+	w.Write([]byte("complex"))
+	printInt(w, int64(floatPrecision*2), 10)
+	r := real(c)
+	w.Write([]byte("("))
+	w.Write([]byte(strconv.FormatFloat(r, 'g', -1, floatPrecision)))
```
sparkprime: why not just call printFloat.  Also, why support complex numbers at atll? :)

sparkprime: Never mind, I see you coded it for the whole Go language pretty much.  It needs tests though because otherwise we have unexercised code in here.

dump/utils.go
```diff
@@ -0,0 +1,97 @@
...
+func isPointerValue(v reflect.Value) bool {
+	switch v.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Slice, reflect.UnsafePointer:
+		return true
+	}
+	return false
+}
+
+func isPrimitivePointer(v reflect.Value) bool {
+	if v.Kind() == reflect.Ptr && isPrimitiveValue(v.Elem()) {
```
sparkprime: What if it's a nil pointer?

dump/pointermap.go
```diff
@@ -0,0 +1,152 @@
...
+type pointerMap struct {
+	allpointers       []uintptr
+	reusedPointers 	  []uintptr
+	primitivePointers []uintptr
+}
+
+// GetPointers: Given a structure, it will recursively map all pointers mentioned in the tree,
+// breaking circular references and provide list of:
+// * all pointers
+// * all pointers that was referenced at least twice
```
sparkprime: that were referenced

dump/pointermap.go
```diff
@@ -0,0 +1,152 @@
...
+			return
+		}
+	}
+	pm.primitivePointers = append(pm.primitivePointers, ptr)
+}
+
+// addPointer to the pointerMap, update reusedPointers. Returns true if pointer was reused
+func (pm *pointerMap) addPointer(ptr uintptr) bool {
+	// Is this already known to be reused?
+	for _, have := range pm.reusedPointers {
```
sparkprime: Does this not cause a performance problem when you write out the AST?

dump/dump.go
```diff
@@ -0,0 +1,510 @@
...
+	"os"
+	"reflect"
+	"regexp"
+	"sort"
+	"strconv"
+)
+
+var packageNameStripperRegexp = regexp.MustCompile("\\b[a-zA-Z_]+[a-zA-Z_0-9]+\\.")
+
+// Options represents configuration options for litter
```
sparkprime: What's litter?

sparkprime:
Using reflection is a bit different to how @sbarzowski was originally envisioning this, he was expecting to write a visitor pattern.  However if what you did works, it has the benefit of automatically supporting new AST types, which is nice.

Can you add some documentation about what exactly is aliased and what isn't?  I believe only primitive pointers are aliased?  Is there any interning of values (i.e. de-duping) beyond what is done in the parser?  It could be dangerous to de-dup things if later passes then modify one, expecting the other to stay the same.

Finally, you did this to improve performance, right?  By how much does it improve performance?

mqliang:
@sparkprime 
>I believe only primitive pointers are aliased?

Reused pointers are also aliased. For example 
```
type Zeo struct {
	A string
}

var z = Zeo {
	A: "hello world",
}

var Obj = &struct {
	Foo *Zeo
	Bar *Zeo
}{
	Foo: &z,
	Bar: &z,
}

```
will be dumped as:
```
var p1 = &Zeo{
	A: "hello world",
}
var Obj = &struct { Foo *Zeo; Bar *Zeo }{
	Foo: p1,
	Bar: p1,
}
```

dump/utils.go
mqliang: It will be parsed as a nil pointer, instead of a pointer pointing to a primitive value.

sparkprime:
> Have not write a performance test yet. But when I run "go test" without pre-parsed stdlib on my laptop, the whole test will take > 10 s; however, it only takes ~0.5 s with pre-parsed stdlib. Is it necessary to write a performance test ?

That sounds good.  Does it degrade performance for other cases, like due to increased binary size or anything?

sparkprime:
> > Is there any interning of values (i.e. de-duping) beyond what is done in the parser? It could be dangerous to de-dup things if later passes then modify one, expecting the other to stay the same.
> Could you please be more specific, I can not quite understand.

As long as you preserve and restore the original aliasing, it is fine.  I'm pretty sure that's what you're doing.

dump/pointermap.go
mqliang: I think it does not matter so much, because even if it cause a performance problem, it's at pre-processing stage, not at runtime stage. Currently "go generate" command takes ~3 seconds on my laptop, which I think is acceptable. And we can optimize it if necessary.

mqliang:
> Does it degrade performance for other cases, like due to increased binary size or anything?

binary size without pre-parsed stdlib: 6.2M
binary size with pre-parsed stdlib: 9.3M

sbarzowski:
> binary size without pre-parsed stdlib: 6.2M
> binary size with pre-parsed stdlib: 9.3M

Looks pretty reasonable to me. And if should go down if we improve desugaring later.

mqliang:
comment addressed. PTAL

dump/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Package dump can dump a Go data structure to Go source file, so that it can be statically embeded into other code.
+
+# Project Status
+* UnsafePointer is not supported yet
+* cyclic pointer is not supported yet
+
+# Implementation Notes
+
+## Aliase
```
sparkprime: Aliasing

dump/README.md
```diff
@@ -0,0 +1,72 @@
+# dump
+
+Package dump can dump a Go data structure to Go source file, so that it can be statically embeded into other code.
+
+# Project Status
+* UnsafePointer is not supported yet
+* cyclic pointer is not supported yet
```
sparkprime: "Cycles in the object graph are not supported yet"

dump/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Implementation Notes
+
+## Aliase
+
+Primitive pointer and reused pointer will be aliased.
```
sparkprime: "Aliases to primitives and structs will be preserved in the objects described by the generated code."

dump/dump_test.go
```diff
@@ -0,0 +1,456 @@
...
+				return 10
+			},
+			"var Obj = 10\n",
+		},
+		{
+			"int8",
+			func() interface{} {
+				return int8(10)
+			},
+			"var Obj = 10\n",
```
sparkprime: The type is changing here.  Is that OK?  You could preserve the type, but if you don't need to, you could just document this instead.

dump/dump_test.go
```diff
@@ -0,0 +1,456 @@
...
+				return float64(10.5)
+			},
+			"var Obj = 10.5\n",
+		},
+		{
+			"complex64",
+			func() interface{} {
+				return complex64(10 + 10.5i)
+			},
+			"var Obj = complex64(10+10.5i)\n",
```
sparkprime: You could just do this for all the primitive types, and then they'd be preserved.  Or am I missing something?

mqliang:
@sparkprime comment addressed. If no further comment, will squash commits.

mqliang:
@sparkprime Any further comments, I will squash the commits if no comments anymore.

sparkprime:
No I think it's good to go.  I'll squash it from here.

PR #102, opened 2017-10-02
Support $ in comprehesions, fixes #68

### LINKED ISSUE #68: Support $ for object comprehensions

sbarzowski:


sbarzowski:
@sparkprime @lwander

desugarer.go
```diff
@@ -236,6 +236,11 @@ func desugarArrayComp(comp *ast.ArrayComp, objLevel int) (ast.Node, error) {
 
 func desugarObjectComp(comp *ast.ObjectComp, objLevel int) (ast.Node, error) {
 
+	if objLevel == 0 {
+		dollar := ast.Identifier("$")
+		comp.Fields = append(comp.Fields, ast.ObjectFieldLocalNoMethod(&dollar, &ast.Self{}))
```
sparkprime: Does this modify the input AST?  Is that OK?

sbarzowski: Yes, it does. It's not a problem, because desugaring in both C++ and Go modifies the original AST instead of creating a new one. In particular we append this special field to regular objects, too.

I think immutable AST could be cleaner, especially for public API, but that's a separate discussion (and out of scope for this change).

PR #105, opened 2017-10-02
Support --ext-str and --ext-code options

sbarzowski:


sbarzowski:
It's necessary for some tests in C++ test suite.

@lwander @sparkprime

PR #107, opened 2017-10-03
Fix tests after rebasing

sbarzowski:


sbarzowski:
@sparkprime 

BTW it would be amazing if we could explicitly rebase and run tests before merging. I had it set up this way with gerrit and jenkins, and it completely solved the problem.

PR #109, opened 2017-10-04
Tailstrict simple

sbarzowski:


sbarzowski:
@sparkprime @lwander

interpreter.go
```diff
@@ -372,7 +389,8 @@ func (i *interpreter) evaluate(a ast.Node) (value, error) {
 				return nil, err
 			}
 			// TODO(https://github.com/google/jsonnet/issues/377): non-integer indexes should be an error
-			return e.evaluate(target.elements[int(indexInt.value)])
+			return e.evaluateTailCall(target.elements[int(indexInt.value)], tc)
```
sparkprime: I think the C++ implementation would use nonTailCall here, but actually this does make sense

interpreter.go
```diff
@@ -415,21 +433,23 @@ func (i *interpreter) evaluate(a ast.Node) (value, error) {
 			bindEnv.upValues[bind.Variable] = th
 		}
 		i.newLocal(vars)
+		sz := len(i.stack.stack)
 		// Add new stack frame, with new thunk for this variable
 		// execute body WRT stack frame.
-		v, err := e.evalInCurrentContext(ast.Body)
-		i.stack.pop()
+		v, err := e.evalInCurrentContext(ast.Body, tc)
+		i.stack.pop(sz)
```
sparkprime: why pop multiple things? can you add a comment

sbarzowski: This is not popping multiple frames. It pops one frame if it's the right frame, i.e. wasn't pushed before. I changed the name of the function to reflect that.

PR #114, opened 2017-10-05
Clean up some TODOs

sbarzowski:
Some were stale, some were transformed into issues, some were fixed


sbarzowski:
@sparkprime @lwander

PR #117, opened 2017-10-06
Eliminate duplication in builtin names

sbarzowski:
Actually there was one (objectFieldEx) that was inconsistent, i.e.
it was available by a different name from what appeared in the stack
trace. This is now fixed.


sbarzowski:
@lwander @sparkprime

PR #119, opened 2017-10-06
Represent types as Go objects, not strings (give them pointer identity)

sbarzowski:
Faster comparisons and IMO cleaner feeling overall.
It doesn't affect behaviour in any way.


sbarzowski:
@lwander @sparkprime

PR #120, opened 2017-10-06
Support for import callbacks

sbarzowski:
It was already there, it was a matter of exposing it in the API


sbarzowski:
@sparkprime @lwander

PR #121, opened 2017-10-09
Some initial test metadata

sbarzowski:
It will be useful for properly testing TLAs, import callbacks etc.


sbarzowski:
@sparkprime @lwander

PR #127, opened 2017-10-10
Post-rebase fix

sbarzowski:
Dead code removed in one branch, ended up being used in another. Probably it's going to be removed anyway in the end when tests are modernized, but for now we need a way to run tests.


sbarzowski:
@sparkprime

PR #129, opened 2017-10-13
Update README.md

sparkprime:
@sbarzowski


README.md
```diff
@@ -9,19 +9,39 @@
...
+{
+   "x": 10,
+   "y": 10
+}
+```
 
 ## Implementation Notes
 
-We are generating some helper classes on types by using http://clipperhouse.github.io/gen/.  Do the following to regenerate these if necessary:
+We are generating some helper classes on types by using
```
sbarzowski: Maybe we should add a note to run `go generate` when std/std.jsonnet changes.

PR #133, opened 2017-10-25
Rename `jsonnet.nativeFunction` -> `jsonnet.NativeFunction`

hausdorff:
NOTE: I have not yet tested this patch, as I cannot seem to run the `tests.sh` file. I am putting it up anyway just to make sure that I'm on at least the right track.

Currently, `jsonnet.VM#NativeFunction` takes a single argument of type
`jsonnet.nativeFunction`. This is ok for internal use, but because this
type is private to the `jsonnet` package, it is not possible for a
third party to call this function (since it can't instantiate the type).

This commit makes this type public to remedy this problem.

### DIFF
--- a/interpreter.go
+++ b/interpreter.go
@@ -217,7 +217,7 @@ type interpreter struct {
 	extVars map[string]potentialValue
 
 	// Native functions
-	nativeFuncs map[string]*nativeFunction
+	nativeFuncs map[string]*NativeFunction
 
 	// A part of std object common to all files
 	baseStd valueObject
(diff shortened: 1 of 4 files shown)


seh:
How does a caller of `(*jsonnet.VM).NativeFunction` acquire a `*jsonnet.NativeFunction`? I don't see any functions that return one. @hausdorff noted that outsiders can't call on `(*jsonnet.VM).NativeFunction` if `jsonnet.NativeFunction` is no exported, but I don't see how exporting it without exporting any of its fields makes it usable. What am I missing here?

seh:
I had missed [discussion in the "ksonnet" channel of the "Kubernetes" Slack team](https://kubernetes.slack.com/archives/C6JLE4L9X/p1508894066000009) on the subject. That describes the problem, in part, but doesn't clarify the solution employed here.

PR #136, opened 2017-10-25
Add Version()

sparkprime:
@hausdorff

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -141,3 +141,7 @@ func snippetToAST(filename string, snippet string) (ast.Node, error) {
 	}
 	return node, nil
 }
+
+func Version() string {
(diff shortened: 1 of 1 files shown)


hausdorff:
LGTM

vm.go
```diff
@@ -141,3 +141,7 @@ func snippetToAST(filename string, snippet string) (ast.Node, error) {
 	}
 	return node, nil
 }
+
+func Version() string {
+	return "v0.9.5"
```
seh: - [ ] Consider passing that in via linker flags.  
  Assuming the existing of `var Version string`:
  _go build -ldflags '-x jsonnet.Version=v0.9.5'_

PR #137, opened 2017-10-29
For compatability, emit the \n within the library, not the cmd

sparkprime:


interpreter.go
```diff
@@ -723,14 +723,14 @@ func serializeJSON(v interface{}, multiline bool, indent string, buf *bytes.Buff
 	}
 }
 
-func (i *interpreter) manifestAndSerializeJSON(trace *TraceElement, v value, multiline bool, indent string) (string, error) {
+func (i *interpreter) manifestAndSerializeJSON(trace *TraceElement, v value, multiline bool, indent string) (*bytes.Buffer, error) {
```
sbarzowski: Maybe it would be better to take this buffer as an argument. Returning a buffer seems weird.

seh: That, or even more generally, demand an `io.Writer`. The downside then is all the potential errors you have to deal with, whereas with a `bytes.Buffer` the operations cannot fail.

sparkprime:
buf now passed in, and I added manifestString into the mix

sbarzowski:
LGTM

PR #139, opened 2017-11-03
String token end location off-by-one error

sparkprime:


sbarzowski:
Nice. Tests need to be updated to reflect this.

sparkprime:
Looks like I broke the tests previously.  I'm not sure why go-tip is failing now though.

PR #146, opened 2017-11-03
Efficient native implementation of std.join

sbarzowski:
It changes the behavior slightly - mixing strings
and arrays in join is no longer allowed.


sbarzowski:
.

sbarzowski:
@sparkprime @lwander

sparkprime:
Changing the behavior is OK.  I don't think one can consider this as useful behavior:

```
$ jsonnet -e 'std.join([],[[1],"2",[3]])'
"[1]2[ ][3]"
```

sparkprime:
I haven't forgotten this btw, I just want to make the incompatible change to both sides simultaneously.

sbarzowski:
FYI I have started working on doing the same thing in C++. I got a bit stuck, though, debugging replacing a frame in a builtin.

sparkprime:
I've changed my mind about doing performance optimizations on both sides. I think we need to encourage people to switch over so let's focus our efforts here.

PR #147, opened 2017-11-08
Fix unparsing of enums, avoid static type error

sparkprime:
@mqliang I'm going to commit this because it fixes the build but please take a look anyway.


sparkprime:
Basic problem: One can't do this:

```
    Fields: DesugaredObjectFields{
        DesugaredObjectField{
            Hide: int(0),
```

One has to do this:

```
    Fields: DesugaredObjectFields{
        DesugaredObjectField{
            Hide: ObjectFieldHide(0),
```

mqliang:
LGTM

PR #148, opened 2017-11-09
Sort FreeVariables slice

mqliang:
Sort FreeVariables slice. Otherwise, every time we regenerate the standard library AST will produce different output, even if "std.jsonnet" keeps unchange.

### DIFF
--- a/ast/util.go
+++ b/ast/util.go
@@ -1,7 +1,41 @@
+/*
+Copyright 2016 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


mqliang:
CI failed, error message is like:
```
/tmp/go-build122298966/b109/desugarer.go:332: Sprintf format %s has arg field.Kind of wrong type github.com/google/go-jsonnet/ast.ObjectFieldKind
```
However, there is no `Sprintf()` statement around desugarer.go:332, don't know what's going wrong

sbarzowski:
Makes sense. This problem is unrelated to the change, the build was failing on Go tip for a while now. I hope this change will fix it: #149

sparkprime:
Interesting problem.  Maybe we should just generate it at build time instead of checking it in.  Does Go tooling make that hard?  Is it bad to have some build.sh that wraps everything?

Otherwise another option is to use datastructures where the importance (or not) of ordering is clear so we can canonicalize them at serialization time rather than canonicalize as we build the datastructures in RAM.  The problem with that is we usually choose datastructures for performance (space / time) rather than so that they look the same in memory every time the program runs :)

PR #149, opened 2017-11-09
Fix bad format string

sbarzowski:


sparkprime:
Looks like moved on to:

```
/tmp/go-build607224492/b108/cmd.go:118: Fprintln call has possible formatting directive %s
```

Which is another location that does not make sense.

Should we be using %v for everything (even strings)?  I figure %v is like Python repr and will quote strings?

sparkprime:
I'll merge it since it improves the situation at least.

PR #150, opened 2017-11-09
remove escLocalFS

mqliang:
escLocalFS is used to embed "std.jsonet" file into jsonnet binary, however with pre-parsed AST it's not needed any more


mqliang:
CI failure should be fixed by https://github.com/google/go-jsonnet/pull/151

PR #152, opened 2017-11-13
Expose MemoryImporter data

kyessenov:
MemoryImporter is convenient to use in an embedded VM setting when an import needs to be passed to a script.

Signed-off-by: Kuat Yessenov <kuat@google.com>


sbarzowski:
Seems like the Travis failure is a result of this https://github.com/mattn/goveralls/issues/93.

sbarzowski:
LGTM @sparkprime

PR #153, opened 2017-11-24
Make fields of ImportedData public

sparkprime:
@sbarzowski @hausdorff @eicnix

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -24,20 +24,23 @@ import (
 )
 
 type ImportedData struct {
-	Err       error
 	FoundHere string
 	Content   string
 }
 
 type Importer interface {
-	Import(codeDir string, importedPath string) *ImportedData
+	Import(codeDir string, importedPath string) (*ImportedData, error)
 }
 
 type ImportCacheValue struct {
+	// nil if we got an error
 	data *ImportedData
 
-	// nil if we have only imported it via importstr
+	// nil if we got an error or have only imported it via importstr
 	asCode potentialValue
+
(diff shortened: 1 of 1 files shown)


imports.go
```diff
@@ -125,22 +125,22 @@ func tryPath(dir, importedPath string) (found bool, content []byte, foundHere st
 func (importer *FileImporter) Import(dir, importedPath string) *ImportedData {
 	found, content, foundHere, err := tryPath(dir, importedPath)
 	if err != nil {
-		return &ImportedData{err: err}
+		return &ImportedData{Err: err}
```
seh: - [ ] Did you consider instead making the function signature return `(*ImportedData, error)` and removing the `Err` field from `ImportedData`?  
  That is, `Import` either returns imported data or an error. When returning an error, the `*ImportedData` value would be nil.

imports.go
```diff
@@ -149,7 +149,7 @@ type MemoryImporter struct {
 
 func (importer *MemoryImporter) Import(dir, importedPath string) *ImportedData {
```
seh: - [ ] Likewise here, what about returning `(*ImportedData, error)` instead?

imports.go
sbarzowski: IIRC I originally put `err` inside, because I wanted errors to be cached. But I agree this feels a bit unnatural.

sparkprime: Have done

seh: Nice. 👍

PR #157, opened 2017-12-08
jsonnet: Fix --output-file command line option

camh-:
The long option --output-file was not working, due to what looks like a
cut and paste error, leaving --exec as the string to test for.


sbarzowski:
LGTM @sparkprime

PR #158, opened 2017-12-08
Native implementation of std.range

sbarzowski:


sparkprime:
Did you have a concrete need to do this?  Or is it just because it's an easy improvement to make?

sbarzowski:
Mostly because it was easy and I wanted to finish something today :-).

But I think it's going to be useful. Slow std.range was a bottleneck (or at least an important factor) in some simple benchmarks I tried. Once it's out of the way it will be easier to measure and optimize other stuff.

Also I think most basic functions, that are likely to affect final performance, should have native implementations eventually. In particular I mean:
- substr
- splitLimit
- slice
- foldl/foldr
- abs
- max/min
- flattenArrays
- sort
- uniq
This is very easy in Go and each can be a separate task, so it may be a good way to utilize time between meetings, classes etc.

Issue #131, opened 2017-10-16
proposal: inject or register error formatter from separate package


kardianos:
github.com/fatih/color is a small dep, but holds a larger dep tree. Coloring output is great for a terminal program, but not so great for a package.

I can work something up if you approve of the idea. I have a google CLA for the golang/go project already.

PR #159, opened 2017-12-08
Remove unsued dependencies

kardianos:
The imports in "_gen.go" were not used, compiled, or
apparently useful so remove.

The color field in the ErrorFormatter was never set
to true and never used. This brought in many unused
dependencies including x/sys/unix that aren't ever
used.

While these aren't huge issues, they add significant
un-needed burden when reviewing imports. I have to
remove them before using the package.

Fixes google/go-jsonnet#131

### DIFF
--- a/error_formatter.go
+++ b/error_formatter.go
@@ -18,25 +18,50 @@ package jsonnet
 import (
 	"bytes"
 	"fmt"
+	"io"
 
-	"github.com/fatih/color"
 	"github.com/google/go-jsonnet/ast"
 	"github.com/google/go-jsonnet/parser"
 )
 
-type ErrorFormatter struct {
-	// MaxStackTraceSize  is the maximum length of stack trace before cropping
-	MaxStackTraceSize int
+type ErrorFormatter interface {
+	// Format static, runtime, and unexpected errors prior to printing them.
+	Format(err error) string
+
+	// Set the the maximum length of stack trace before cropping.
+	SetMaxStackTraceSize(size int)
+
+	// Set the color formatter for the location color.
+	SetColorFormatter(color ColorFormatter)
+}
+
+type ColorFormatter func(w io.Writer, f string, a ...interface{}) (n int, err error)
+
+var _ ErrorFormatter = &termErrorFormatter{}
+
+type termErrorFormatter struct {
+	// maxStackTraceSize  is the maximum length of stack trace before cropping
+	maxStackTraceSize int
 
 	// Examples of current state of the art.
 	// http://elm-lang.org/blog/compiler-errors-for-humans
 	// https://clang.llvm.org/diagnostics.html
-	pretty   bool
-	colorful bool
-	SP       *ast.SourceProvider
+	color  ColorFormatter
+	pretty bool
+
+	// sp is currently never set, but is used to format locations.
+	sp *ast.SourceProvider
 }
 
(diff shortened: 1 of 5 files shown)


sparkprime:
Colorized output was something we had intended to make available by default.  However it's not worth having a huge dependency on.  Can we not just replace it with a terminal escape code like "\e[0;31m" ?

@sbarzowski

kardianos:
Colorized output would be great for the cmd. Less so for a pkg. I could do a follow on PR or modify the existing one that uses colorfy just in the cmd.

sbarzowski:
> Can we not just replace it with a terminal escape code like "\e[0;31m" ?

Probably. From what I see the main selling points of fatih/color are automagically checking if the output should be colorized (checking if it's a tty) and Windows support. I think we could handle that.

> I could do a follow on PR or modify the existing one that uses colorfy just in the cmd.

:+1:

kardianos:
This PR now enables color output in the jsonnet cmd. Prior to this PR color was present in the code but never enabled. The color code injection happens in the jsonnet cmd. This separates out the color dependency from the rest of the code.  I'm not very happy with the result, esp how SetColorFormatter is part of an interface that really shouldn't be there. But it would be trivial to refactor if another formatter is needed, such as an HTML formatter or JSON formatter, the color could be injeced through setting a global var on the jsonnet package, or putting it into a separate formatter package, but all that is probably over design. I'm straining the limits of over design already as is.

error_formatter.go
```diff
@@ -18,25 +18,50 @@ package jsonnet
...
-	MaxStackTraceSize int
+type ErrorFormatter interface {
+	// Format static, runtime, and unexpected errors prior to printing them.
+	Format(err error) string
+
+	// Set the the maximum length of stack trace before cropping.
+	SetMaxStackTrackeSize(size int)
+
+	// Set the color formatter for the location color.
+	SetColorFormatter(color ColorFormatter)
```
sbarzowski: hmmm... it supports only one color (which is all we're using at the moment)...

error_formatter.go
```diff
@@ -18,25 +18,50 @@ package jsonnet
...
 	// Examples of current state of the art.
 	// http://elm-lang.org/blog/compiler-errors-for-humans
 	// https://clang.llvm.org/diagnostics.html
-	pretty   bool
-	colorful bool
-	SP       *ast.SourceProvider
+	color  ColorFormatter
+	pretty bool
+
+	// sp is currently never set, but is used to format locations.
```
sbarzowski: This may be very confusing. We could replace SourceProvider methods with ordinary functions. The original idea was to keep the source code in SourceProvider, but in the end the reference to the code is kept in each LocationRange anyway... But I think we can leave it as-is for now.

error_formatter.go
```diff
@@ -18,25 +18,50 @@ package jsonnet
...
 
-type ErrorFormatter struct {
-	// MaxStackTraceSize  is the maximum length of stack trace before cropping
-	MaxStackTraceSize int
+type ErrorFormatter interface {
+	// Format static, runtime, and unexpected errors prior to printing them.
+	Format(err error) string
+
+	// Set the the maximum length of stack trace before cropping.
+	SetMaxStackTrackeSize(size int)
```
sbarzowski: typo

kardianos: Done.

PR #161, opened 2017-12-11
Basic refactoring - towards shared e2e tests

sbarzowski:
I would like to be able to run go testsuite on C++ codebase. To do so I need to decouple the tests from the current implementation. This change is the first step.

### DIFF
--- a/jsonnet_test.go
+++ b/jsonnet_test.go
@@ -1,0 +1,118 @@
+package jsonnet
+
+import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


sparkprime:
You just moved some functions around, right?

sbarzowski:
Primarily I split TestMain function into smaller ones. And I moved tests of implementation details to a new file (jsonnet_test.go). There's no change in behavior.

Issue #160, opened 2017-12-10
Incorrect method documentation on EvaluateSnippetMulti


kardianos:
EvaluateSnippetMulti returns an "object" `map[string]string`, not an array.

sbarzowski:
Thanks for noticing this. It seems like the whole doc for `EvaluateSnippetMulti` is copy-pasted form `EvaluateSnippetStream`.

PR #163, opened 2017-12-13
jsonnet: fix docs, correct EvaluateSnippetMulti

kardianos:
EvaluateSnippetMulti comment was copied from another function by mistake.
Add a space between the license and the package name.
Add a few periods after the comments.

Fixes google/go-jsonnet#160

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -83,7 +83,7 @@ func (vm *VM) TLACode(key string, val string) {
 	vm.tla[key] = vmExt{value: val, isCode: true}
 }
 
-// Importer sets Importer to use during evaluation (import callback)
+// Importer sets Importer to use during evaluation (import callback).
 func (vm *VM) Importer(i Importer) {
 	vm.importer = i
 }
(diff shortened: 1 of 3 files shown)


vm.go
```diff
@@ -151,8 +151,8 @@ func (vm *VM) EvaluateSnippetStream(filename string, snippet string) (docs []str
 	return
 }
 
-// EvaluateSnippetStream evaluates a string containing Jsonnet code to an array.
-// The array is returned as an array of JSON strings.
+// EvaluateSnippetMulti evaluates a string containing Jsonnet code to key-value
+// pairs. The keys and values are both returned as JSON strings.
```
sbarzowski: "The keys and values are both returned as JSON strings." sounds like it returned an an array of pairs of pair of arrays of strings.  And it actually returns a map[string]string. And while values in this map are JSONs, keys are not.

kardianos: That was careless of me. Fixed.

sparkprime:
Thanks

PR #165, opened 2017-12-17
Make errors in native functions runtime errors

sbarzowski:
Instead of crashing with internal error

PR #166, opened 2017-12-17
Add test case for native function returning error

camh-:
A native function returning an error should be reported as a runtime
error, with a stack trace.


camh-:
The test in this PR will fail without the change in #165.

PR #167, opened 2017-12-23
Basic support for running jsonnet command in tests

sbarzowski:
A step towards a unified test suite. In particular this
lets us use C++ version with Go tests. The problem remains
that errors (both static and runtime) may differ between implementations.
This will require special handling. C++ version seems to pass all
"positive" tests. Go version also has this problem (error formatter used in tests is different).

Also native functions etc. are not handled in any way at the moment.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"flag"
 	"io/ioutil"
+	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


sparkprime:
I wonder whether it would be easier to make the errors align or to make the tests be flexible about differences in error messages :)  The easiest option is just to have separate golden files for both side, but that is ongoing maintenance pain.

PR #169, opened 2018-01-10
Call top level function even if there are no TLAs (match cpp semantics)

### LINKED ISSUE #168: RUNTIME ERROR: Couldn't manifest function in JSON output.
I found this in code:
```
case *valueFunction:
		return nil, makeRuntimeError("Couldn't manifest function in JSON output.", i.getCurrentStackTrace(trace))
```
and my input is something like this:
```
local a = import "../../../applib/a.libsonnet";
local b = import "../../../applib/b.libsonnet";

function(config={})
  local overall_config = default_config + config;
  (if a.match(overall_config) then
    {
      ...
    }
  else
    {
      ...
    }) + {
    ...
  }
```
And i get this:
```
RUNTIME ERROR: Couldn't manifest function in JSON output.
	During manifestation
```
How can i resolve this?
Thanks!

sparkprime:
Fixes #168 

@sbarzowski


sbarzowski:
@sparkprime there is a test that fails now (a function with no arguments) that verified the old behavior.  The output needs to be regenerated. It would also make sense to add two more tests:
* `function(x) 42` - to check what happens when there is a missing argument for the top level function
* `{ f(x):: 42 }` - to check the "Couldn't manifest function in JSON"

PR #176, opened 2018-01-15
Very basic, experimental linter

sbarzowski:
No tests yet. It has currently one feature - reporting unused variables. Internally it keeps all the usages. Expect rough edges. I have only run it on a few files by hand so far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@ coverage.out
 .*.swp
 /tests_path.source
 /jsonnet/jsonnet
+/linter/jsonnet-lint/jsonnet-lint
 *.so
 *.prof
--- a/linter/README.md
+++ b/linter/README.md
@@ -1,0 +1,10 @@
+# jsonnet-lint
+
+Very basic experimental, quick and a bit dirty linter. It's intended to become a full featured linter as it matures. PRs are very welcome!
+
+Don't expect much from it now.
+
+## Notes
+
+It currently runs on a desugared file. It has its good and bad sides.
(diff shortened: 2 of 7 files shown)


sbarzowski:
@sparkprime

sparkprime:
Nice. The only thing that makes this not immediately useful is that sometimes function params are deliberately unused because you're fitting into a prototype defined by someone else. I don't think we have a way out of that right now because we can't insist they be renamed to `unused_x` now that we have named argument passing. So, we need some comment to mark it as unused / disable the lint warning on that line *but* we've lost all comments in the parsing phase.  Still, it's a good start and I'm inclined to merge it as-is.

sbarzowski:
> The only thing that makes this not immediately useful is that sometimes function params are deliberately unused because you're fitting into a prototype defined by someone else.

For this reason it doesn't issue any warnings about unused params. Only local variables. Here's an if that checks it:
https://github.com/google/go-jsonnet/pull/176/files#diff-80f9fdcc81aed508ba78619f9dd78083R47

sparkprime:
Here's some food for thought http://pylint-messages.wikidot.com/all-codes

Some stuff that might be immediately useful:
- check printf format strings
- warn if the same thing imported twice
- disallow imports except at the top (maybe)
- basic static analysis for function signatures, check args match
- anything the reformatter would fix (or modify the reformatter to give errors instead of fixing).

sparkprime:
Oh yeah I missed that, heh.  Ok let's just merge it then.

Issue #132, opened 2017-10-17
Weird std.exp(something_big) on some platforms.


sbarzowski:
We're affected by this: https://github.com/golang/go/issues/14932

BTW it seems it is non-trivial to have exactly the same values of numeric functions on all platforms - https://github.com/golang/go/issues/20319.

sparkprime:
There's a case for removing (a)sin, (a)cos, (a)tan, sqrt, exp, log, pow for precisely this reason.

We can keep floor, ceil, exponent, mantissa as they are straightforward and should be the same everywhere.

We could provide an integer version of log -- one that returns ceil(log(floor(a), floor(b))).  We can provide an integer version of pow via a ** operator.

sbarzowski:
I don't think there is any action is going to be needed here, so I'll close this. We have some tests which try to ensure that floating point operations behave reasonably in general.

PR #178, opened 2018-01-16
Oldest version known to work = Go 1.8, closes #177

### LINKED ISSUE #177: [BUG] compiler crash using go 1.6.x
hi, it's me again and again....

## how to reproduce
I tried to compile jsonnet on a brand new vm on gce, using the ubuntu 16.04 and n1-highcpu-2（2  vCPU,1.8 GB）
I downloaded the go binary from https://dl.google.com/go/go1.6.4.linux-amd64.tar.gz and install following the instruction at https://golang.org/doc/install.
And i executed `go get -u github.com/fatih/color`, then `go build`.

image

image

image

sbarzowski:
It seems like it was broken on older compilers
since pregenerated stdlib AST.

Add go 1.8.x to travis, so that we have a chance
of catching when our "known to work" no longer holds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 sudo: false
 go:
   - 1.x
+  - 1.8.x
   - tip
 before_install:
   - go get github.com/axw/gocov/gocov
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ feature complete but is not as heavily exercised as the [Jsonnet C++
 implementation](https://github.com/google/jsonnet).  Please try it out and give
 feedback.
 
-This code is known to work on Go 1.6 and above.
+This code is known to work on Go 1.8 and above. We recommend always using the newest stable release of Go.
 
 ## Build instructions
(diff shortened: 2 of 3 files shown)


sbarzowski:
@sparkprime

sbarzowski:
hmmm... It seems golang changed the behavior of math.Exp. This caused problems before: #132 

Not sure what to do now about it.

sparkprime:
How about modifying the test to check rather than a == b that std.abs((a / b) - 1) < 0.0000001 or something like that

sbarzowski:
Wouldn't help, the problem is 0 vs inf when running `std.exp(100000000000000000000)`. Go docs actually say:

> Very large values overflow to 0 or +Inf.

Seems a bit silly to me. But I guess I can check for that.

sparkprime:
Tbh you can probably just remove that test, it's not like it's a realistic thing to do

sbarzowski:
@sparkprime ping

PR #181, opened 2018-01-17
Convert non-string errors to string

sbarzowski:


interpreter.go
```diff
@@ -364,6 +364,12 @@ func (i *interpreter) evaluate(a ast.Node, tc tailCallStatus) (value, error) {
 			// error when evaluating error message
 			return nil, err
 		}
+		if msgVal.getType() != stringType {
+			msgVal, err = builtinToString(e, &readyValue{msgVal})
```
sbarzowski: Stack traces are a little problematic as usual. 

Calling builtinToString directly would be normally a bad thing (the right way would be to get a functionValue for builtinToString, call it, and evaluate the call - this way it would appear in the stack trace etc.). The problem is that it would not show a stack trace line for the error statement.

PR #183, opened 2018-01-27
Fix processing the same file many times

sbarzowski:
There was a trivial bug, which disabled caching
of parsed, desugared and analyzed code. So imports
were only cached on data level, not on executable
ast level.

This change will likely result in ~10x speedup
for some users.


sbarzowski:
@sparkprime @redbaron

PR #185, opened 2018-01-30
Convert --tla/ext-*-file to imports

camh-:
Instead of reading the files directly for --tla/ext-*-file, just add a
code snippet that does an import or importstr.

This makes relative imports work when the code-file path is not in the
current directory. It also makes the import lazy so if the arg/var is
not referenced, the file does not have to exist.

Fixes: google/jsonnet#452


sparkprime:
Thanks for doing both implementations and adding a test too!

PR #186, opened 2018-01-31
Fast, native implementation of strReplace

sbarzowski:


sbarzowski:
Inspired by https://github.com/google/jsonnet/pull/454

On testdata/strReplace.jsonnet it goes from 20s to 0.02s (that's 1000x).

@sparkprime

gburiola:
This is great. Thanks for working on it @sbarzowski 

Out of curiosity, is there any reason for creating new `stdlib` functions in jsonnet instead of native ?

sbarzowski:
> is there any reason for creating new stdlib functions in jsonnet instead of native?

Yes, there are many:
a) You have it in both C++ jsonnet and go-jsonnet, with builtin implementations you need to do it twice.
b) Even though writing builtins in go-jsonnet, it's often easier to do in jsonnet. You can play with it without modifying the interpreter - faster iterations. And it's enough to know jsonnet to contribute and understand.
c) Implementation in jsonnet is a bit like a specification and being able to implement it reasonably in jsonnet is a good sign that it can be a good builtin with well-defined behavior.
d) You can start using it immediately without waiting for it to be accepted. And if the proposal doesn't end up in the stdlib for whatever reason, it can still be used as a library.

So I think new additions to standard library should have a jsonnet implementation and native implementation can be added later if needed.

PR #188, opened 2018-02-16
Implement extvar/tla file via code import

sparkprime:
Equivalent of https://github.com/google/jsonnet/pull/455/files#diff-cbfdcc436b29a37aa318778257ca6871


sparkprime:
Hah, already done...

PR #191, opened 2018-02-17
Do not assume that import path is always a filesystem path

### LINKED ISSUE #190: Import path should be opaque
When importing jsonnet documents using the `import` statement, the current implementation assumes that the resource returned from importer's `FoundHere` is a filesystem path. On subsequent imports, the base dir of that path is derived via a call to `path.Dir()` (see e.g. https://github.com/google/go-jsonnet/blob/master/interpreter.go#L415). This function always performs a path cleanup.

This is a problem when implementing custom importers that do not use filesystem paths, but for example URLs. E.g. when an importer returns `https://raw.githubusercontent.com/ksonnet/ksonnet-lib/master/ksonnet.beta.2/k.libsonnet` in `FoundHere`, upon resolving relative imports based on that URL, the `importDir` comes back to the importer cleaned up as `https:/raw.githubusercontent.com/ksonnet/ksonnet-lib/master/ksonnet.beta.2` (notice the single forward slash in `https:/`).

`go-jsonnet` should not assume filesystem paths, in fact the resource "path" should be opaque.

petr-k:
Closes #190

### DIFF
--- a/interpreter.go
+++ b/interpreter.go
@@ -412,11 +412,11 @@ func (i *interpreter) evaluate(a ast.Node, tc tailCallStatus) (value, error) {
 		return nil, e.Error(fmt.Sprintf("Value non indexable: %v", reflect.TypeOf(targetValue)))
 
 	case *ast.Import:
-		codeDir := path.Dir(node.Loc().FileName)
+		codeDir, _ := path.Split(node.Loc().FileName)
 		return i.importCache.ImportCode(codeDir, node.File.Value, e)
 
 	case *ast.ImportStr:
-		codeDir := path.Dir(node.Loc().FileName)
+		codeDir, _ := path.Split(node.Loc().FileName)
 		return i.importCache.ImportString(codeDir, node.File.Value, e)
 
 	case *ast.LiteralBoolean:


petr-k:
I signed it!

sparkprime:
On Windows, will this try to split a URI with `\` instead of `/` ?

sparkprime:
I just checked the C++ implementation which always just uses `/` however that probably means it is broken for things like `import @"c:\foo.jsonnet"`.  Tricky one, this!

petr-k:
@sparkprime No, just as the original `path.Dir`, this uses the `path` package, which deals only with forward-slash delimited paths.

petr-k:
@sparkprime Agreed that it is probably broken on Windows (and has been in C++ jsonnet too), but I propose that a separate issue is logged for that. What do you think?

sbarzowski:
I think this is a fine short-term solution. Long-term I think since we allow custom importers we shouldn't assume anything about the path (abstract any splitting and/or getting absolute path to importer).

anguslees:
@sparkprime: I think long term, we should keep file paths as completely opaque strings to the core go-jsonnet code: Change `Importer.Import` to provide the "imported from" arg unmolested, rather than just the dir component.  It would then be up to the specific importer to call `Dir()`, if relevant.

PR #192, opened 2018-02-18
Better Importer interface

sbarzowski:
As discussed in https://github.com/google/go-jsonnet/issues/190

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -23,67 +23,87 @@ import (
 	"path"
 )
 
-// ImportedData represents imported data and where it came from.
-type ImportedData struct {
-	FoundHere string
-	Content   string
-}
-
 // An Importer imports data from a path.
+// TODO(sbarzowski) caching of errors (may require breaking changes)
 type Importer interface {
-	Import(codeDir string, importedPath string) (*ImportedData, error)
+	// Import fetches data from a given path. It may be relative
+	// to the file where we do the import. What "relative path"
+	// means depends on the importer.
+	//
+	// It is required that:
+	// a) for given (importedFrom, importedPath) the same
+	//    (contents, foundAt) are returned on subsequent calls.
+	// b) for given foundAt, the contents are always the same
+	//
+	// It is recommended that if there are multiple locations that
+	// need to be probed (e.g. relative + multiple library paths)
+	// then all results of all attempts will be cached separately,
+	// both nonexistence and contents of existing ones.
+	// FileImporter may serve as an example.
+	Import(importedFrom, importedPath string) (contents Contents, foundAt string, err error)
 }
 
-// ImportCacheValue represents a value in an imported-data cache.
-type ImportCacheValue struct {
-	// nil if we got an error
-	data *ImportedData
-
-	// nil if we got an error or have only imported it via importstr
-	asCode potentialValue
-
-	// Errors can occur during import, we have to cache these too.
-	err error
+// Contents is a representation of imported data. It is a simple
+// string wrapper, which makes it easier to enforce the caching policy.
+type Contents struct {
+	data *string
 }
 
-type importCacheKey struct {
-	dir          string
-	importedPath string
+func (c Contents) String() string {
+	return *c.data
 }
 
-type importCacheMap map[importCacheKey]*ImportCacheValue
+// MakeContents creates Contents from a string.
+func MakeContents(s string) Contents {
+	return Contents{
+		data: &s,
+	}
+}
 
 // ImportCache represents a cache of imported data.
+//
+// While the user-defined Importer implementations
+// are required to cache file content, this cache
+// is an additional layer of optimization that caches values
+// (i.e. the result of executing the file content).
+// It also verifies that the content pointer is the same for two foundAt values.
 type ImportCache struct {
-	cache    importCacheMap
-	importer Importer
+	foundAtVerification map[string]Contents
+	codeCache           map[string]potentialValue
+	importer            Importer
 }
 
-// MakeImportCache creates and ImportCache using an importer.
+// MakeImportCache creates an ImportCache using an Importer.
 func MakeImportCache(importer Importer) *ImportCache {
-	return &ImportCache{importer: importer, cache: make(importCacheMap)}
+	return &ImportCache{
+		importer:            importer,
+		foundAtVerification: make(map[string]Contents),
+		codeCache:           make(map[string]potentialValue),
+	}
 }
 
-func (cache *ImportCache) importData(key importCacheKey) *ImportCacheValue {
-	if cached, ok := cache.cache[key]; ok {
-		return cached
+func (cache *ImportCache) importData(importedFrom, importedPath string) (contents Contents, foundAt string, err error) {
+	contents, foundAt, err = cache.importer.Import(importedFrom, importedPath)
+	if err != nil {
+		return Contents{}, "", err
 	}
-	data, err := cache.importer.Import(key.dir, key.importedPath)
-	cached := &ImportCacheValue{
-		data: data,
-		err:  err,
+	if cached, importedBefore := cache.foundAtVerification[foundAt]; importedBefore {
+		if cached != contents {
+			panic(fmt.Sprintf("importer problem: a different instance of Contents returned when importing %#v again", foundAt))
+		}
+	} else {
+		cache.foundAtVerification[foundAt] = contents
 	}
-	cache.cache[key] = cached
-	return cached
+	return
 }
 
 // ImportString imports a string, caches it and then returns it.
-func (cache *ImportCache) ImportString(codeDir, importedPath string, e *evaluator) (*valueString, error) {
-	cached := cache.importData(importCacheKey{codeDir, importedPath})
-	if cached.err != nil {
-		return nil, e.Error(cached.err.Error())
+func (cache *ImportCache) ImportString(importedFrom, importedPath string, e *evaluator) (*valueString, error) {
+	data, _, err := cache.importData(importedFrom, importedPath)
+	if err != nil {
+		return nil, e.Error(err.Error())
 	}
-	return makeValueString(cached.data.Content), nil
+	return makeValueString(data.String()), nil
 }
 
 func codeToPV(e *evaluator, filename string, code string) potentialValue {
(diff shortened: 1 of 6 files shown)


sbarzowski:
@petr-k @sparkprime @anguslees

I haven't tested it beyond existing test suite, yet.

anguslees:
I don't like the separation/race between canonicalise and import.  Apart from regular race conditions, in my case I want to fetch from remote servers - and it is _much_ more efficient to fetch content in the same operation as walking the search path (testing if the content actually exists).

Can we roll the two back into one `Import` operation, like before this PR?  I think we can just document that it is important that `FoundHere` is in an absolute/canonicalised form.

sbarzowski:
@anguslees 
If you're using urls to access remote servers, then Canonicalize can be a no-op, the paths are already canonical, i.e. it can be:
```
func (importer *RemoteImporter) Canonicalize(codePath, importedPath string) (string, error) {
	return importedPath, nil
}
```

The only requirement for canonical path is that importer knows how to fetch it, there is no need to check that it actually exists.

In case of file importer, I'm aware of the race, but it is completely benign. Jsonnet has every right to show an error if an imported file is deleted - and that's the only thing that can happen. And there we need to check for existence in a bunch of locations anyway.

sparkprime:
Don't we want the ability for a Jsonnet file hosted at http://foo/bar.jsonnet to do `import "baz.jsonnet"` and for that to be resolved to http://foo/baz.jsonnet ?

sparkprime:
I wonder if we could let the importer do cacheing by passing it the cache singleton to use.  I also think maybe we should cache the non-existence of files, to speed up future searches for the same thing.

sbarzowski:
> Don't we want the ability for a Jsonnet file hosted at http://foo/bar.jsonnet to do import "baz.jsonnet" and for that to be resolved to http://foo/baz.jsonnet ?

It doesn't change anything.  The redirections can be completely transparent.  In your example `http://foo/bar.jsonnet` doesn't depend on from where it is imported, so it is already a good canonical path.

Perhaps "canonicalize" gives a wrong idea that it needs to be the sole identifier for given content. Maybe `getKey` or something would be a better name. The only point is that it no longer depends on from where it was imported.

> I wonder if we could let the importer do cacheing by passing it the cache singleton to use.

I think it wouldn't help. It would still need to first get some key to the cache, then check it, then fetch the data. 

> I also think maybe we should cache the non-existence of files, to speed up future searches for the same thing.

It is something the importer could do internally. I see also other, potentially easier options - like for example caching JPATH lookups separately (still within the importer). Anyway it doesn't change the interface.

anguslees:
I think I'm missing something in this discussion:

> The only requirement for canonical path is that importer knows how to fetch it, there is no need to check that it actually exists.

So let's walk through an example:

Let's say that `http://foo/bar.jsonnet` contains `import "baz.libsonnet"`.  I think this means:
- Canonicalize("http://foo/bar.jsonnet", "baz.libsonnet") will presumably return `"http://foo/baz.libsonnet"`
- check cache, missing
- `Import("http://foo/baz.libsonnet")` is expected to return the content.

Now the above works fine for simple direct imports, but what if `baz.libsonnet` needs to be found along the search path?  I need to go and probe for existence, which is expensive.

I _can_ of course avoid the second fetch by pre-stuffing the result in a cache during Canonicalize, and making Import always hit this cache - but this seems like I'm working around (not with) this new proposed API.  In particular, with the workaround Canonicalize now has side effects, while Import doesn't :/

sparkprime:
I don't see how

```
func (importer *RemoteImporter) Canonicalize(codePath, importedPath string) (string, error) {
	return importedPath, nil
}
```

could ever work if the importedPath is relative, since it would collide with the result of Canonicalize for different files (ones with different content).

sbarzowski:
> I don't see how [...] could ever work if the importedPath is relative...

Ok, I assumed that if you use URLs then you always use an *absolute* one, hence some misunderstandings. Of course if you want to have imports relative to the source *and* search path for remote stuff, then it gets more complicated.

sbarzowski:
Hmmm... I'm getting more and more convinced that we should just leave the caching of file contents to the importer, since we're talking about having some sort of cache in the importer anyway. It's also very flexible and it would keep the interface very small.

So now, my plan is to:
1) Change the Importer interface to `Import(from, path) -> (contents, foundAt, err)`. The first argument (`from`) would be a path to the file, not its directory like it is now (to satisfy #190).
2) Decouple code cache and data (file contents) cache. The code cache would use `foundAt` as key, so that each file would be parsed and analyzed exactly once.
3) The Importer wouldn't care about code cache or any Jsonnet data structures.
4) It would be recommended that the Importer implements cache, but it will be an implementation detail, the rest of the code won't care about it.

Does it make sense?

sparkprime:
Yes except we should encourage people to use the cache, otherwise it might expose the laziness.
If you can keep a simple key/value store in core that makes it trivial for the importer to do it, that would be awesome.

anguslees:
Sgtm.

anguslees:
Oh, I encourage you to break the exisiting function signature in some way to force users to reevaluate their use of this API as part of the upgrade.  (ie don't just change the semantics of the first string arg)

sparkprime:
I think we're on first name terms with all of the users of this API so this should be OK :)

Quentin-M:
I just briefly asked @sparkprime in Slack if there is reason `MakeImportCache` is exported if it doesn't implement the `Importer` interface? 

It seems that the VM is very stateless at this point, but it would be very beneficial to have the imported libraries cached across multiple `evaluate` calls. When working with libraries such as [ksonnet-lib](https://github.com/ksonnet/ksonnet-lib), it takes ~1.9s on an Intel i7 (Kaby Lake) @ 3.1 GHz, just to evaluate the following snippet. Those quickly add up when generating several files and it would be silly to use create a snippet that aggregate those files to do an evaluate multi for the purpose of caching the libraries, making the operator loose 1/ ability to debug the snippets easily 2/ ability to pass different ExtVar/ExtCode/TLA to those snippets 3/ file-grained control over what gets executed and when 4/ etc.

```
local k = import "k.libsonnet";
k.core.v1.list.new([])
```

I feel it would be reasonable to cache those in the VM/Importer.
I do not thing that the libraries would usually change during the lifespan of a VM.
If they do, the operator is most-likely aware about it, and a new VM/Importer instance can be created.

sbarzowski:
@Quentin-M 
Thanks, these are good points.

What takes time when importing `k.libsonnet` is parsing and analyzing it, so code (ast) cache is the important part. Importer will only have bytes-level cache (doesn't know about the code), so further changes will be needed.

sbarzowski:
@sparkprime I've updated it. Now `Importer` is responsible for data-level caching. The file cache now caches existence/contents of every tried absolute path.

imports.go
```diff
@@ -30,60 +30,66 @@ type ImportedData struct {
...
 // ImportCache represents a cache of imported data.
 type ImportCache struct {
-	cache    importCacheMap
-	importer Importer
+	foundAtVerification map[string]*string
+	codeCache           map[string]potentialValue
+	importer            Importer
 }
 
 // MakeImportCache creates and ImportCache using an importer.
```
sparkprime: creates an ImportCache

imports.go
```diff
@@ -100,69 +106,101 @@ func codeToPV(e *evaluator, filename string, code string) potentialValue {
...
 
 // MemoryImporter "imports" data from an in-memory map.
 type MemoryImporter struct {
-	Data map[string]string
+	Data map[string]*string
 }
 
-// Import imports a map entry.
-func (importer *MemoryImporter) Import(dir, importedPath string) (*ImportedData, error) {
+// Import fetches data from a map entry.
```
sparkprime: Can you add to comment that codePath is the string describing the file to import, and importedPath is the foundAt of the file doing the importing.

imports.go
sparkprime: also capitalize importer

imports.go
```diff
@@ -30,60 +30,66 @@ type ImportedData struct {
...
+		importer:            importer,
+		foundAtVerification: make(map[string]*string),
+		codeCache:           make(map[string]potentialValue),
+	}
 }
 
-func (cache *ImportCache) importData(key importCacheKey) *ImportCacheValue {
-	if cached, ok := cache.cache[key]; ok {
-		return cached
+func (cache *ImportCache) importData(codePath, importedPath string) (contents *string, foundAt string, err error) {
```
sparkprime: Can you be consistent with the parameter names, because in Importer they are `importedFrom, path`

imports.go
```diff
@@ -30,60 +30,66 @@ type ImportedData struct {
...
-}
-
-type importCacheKey struct {
-	dir          string
-	importedPath string
-}
-
-type importCacheMap map[importCacheKey]*ImportCacheValue
-
 // ImportCache represents a cache of imported data.
```
sparkprime: Please add:

While the user-defined Importer implementations are required to cache file content, this cache is an additional layer of optimization that caches values (i.e. the result of executing the file content).  It also verifies that the content pointer is the same for two foundAt values.

imports.go
```diff
@@ -30,60 +30,66 @@ type ImportedData struct {
...
+	// Import fetches data from a given path. It may be relative
+	// to the file where we do the import. What "relative path"
+	// means depends on the importer.
+	//
+	// It is required that:
+	// a) for given (from, path) the same (contents, foundAt) are
+	//    returned on subsequent calls.
+	// b) for given foundAt, the contents are always the same
+	//
+	// Note that by "the same contents" we mean the same pointer and
```
sparkprime: It is a little bit weird (if you haven't read the implementation) to have to return the same pointer.  We could alternatively say that "If you return the same foundAt for multiple imports, we use the content for the first one."  That would mean cacheing the LiteralString in the codeCache.

On the other hand, it's not a big deal for implementers to actually meet these requirements, so I'm happy either way.

imports.go
```diff
@@ -100,69 +106,101 @@ func codeToPV(e *evaluator, filename string, code string) potentialValue {
...
 // MemoryImporter "imports" data from an in-memory map.
 type MemoryImporter struct {
-	Data map[string]string
+	Data map[string]*string
 }
 
-// Import imports a map entry.
-func (importer *MemoryImporter) Import(dir, importedPath string) (*ImportedData, error) {
+// Import fetches data from a map entry.
+func (importer *MemoryImporter) Import(codePath, importedPath string) (contents *string, foundAt string, err error) {
```
sparkprime: Ditto about parameter naming

imports.go
```diff
@@ -100,69 +106,101 @@ func codeToPV(e *evaluator, filename string, code string) potentialValue {
...
 	}
-	return true, content, absPath, err
+	return entry.exists, &entry.contents, absPath, nil
 }
 
-// Import imports a file.
-func (importer *FileImporter) Import(dir, importedPath string) (*ImportedData, error) {
-	found, content, foundHere, err := tryPath(dir, importedPath)
+// Import imports file from the filesystem.
+func (importer *FileImporter) Import(codePath, importedPath string) (contents *string, foundAt string, err error) {
```
sparkprime: Ditto about parameter naming

imports.go
```diff
@@ -100,69 +106,101 @@ func codeToPV(e *evaluator, filename string, code string) potentialValue {
...
+	var entry *fsCacheEntry
+	if cacheEntry, isCached := importer.fsCache[absPath]; isCached {
+		entry = cacheEntry
+	} else {
+		contentBytes, err := ioutil.ReadFile(absPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				entry = &fsCacheEntry{
+					exists:   false,
+					contents: "",
```
sparkprime: Could omit this line

imports.go
sparkprime: The semantics of this are a bit weird now I think -- all paths are treated as absolute keys into the "Data" map.  That probably should be documented, although I understand we only use this for tests so it does not matter?

sparkprime:
LGTM, I'd like to hear @anguslees opinions.

@Quentin-M This does not address your request, but that is strictly orthogonal to this (a case of re-using the FileImporter between VM calls).  We should do it in a separate PR.

sbarzowski:
@sparkprime I have just applied the suggestions and I changed the interface slightly (`Contents` struct) to catch some caching problems. This way it's clear what it means that "the contents" are the same, the user has no way to change the underlying string. The same could be done with builtin string (just a string, not a pointer), but to have guaranteed O(1) comparison [here](https://github.com/google/go-jsonnet/pull/192/files#diff-0775b42bb390154aa78ab5bd812680c8R91), I would need to use reflection to get StringHeader, so I concluded that a wrapper is less ugly.

sparkprime:
LGTM @anguslees anything to say here?  I'll merge if not.

PR #194, opened 2018-02-21
Add std.sign and fix division by zero in stdlib

sbarzowski:


sbarzowski:
@sparkprime @benley

In response to https://github.com/google/jsonnet/issues/458.

I will send a PR to C++ jsonnet, once it gets accepted here.

PR #196, opened 2018-02-24
Apply golint recommendations

marcelocantos:


parser/lexer.go
```diff
@@ -165,7 +165,8 @@ type token struct {
 	loc ast.LocationRange
 }
 
-type tokens []token
+// Tokens is a slice of token structs.
+type Tokens []token
```
sbarzowski: Any reason for this to be exported?

sbarzowski:
That's very helpful, thanks!

parser/lexer.go
marcelocantos:
```bash
$ golint -min_confidence 0
⋮
lexer.go:685:36: exported func Lex returns unexported type parser.tokens, which can be annoying to use
```

marcelocantos:
I did this PR to clean up my VSCode window. Now that I've run golint on the command line, I realised that there's a bunch of warnings outside the `parser` directory that VSCode wasn't picking up. I'll add fixes for these other warnings shortly.

marcelocantos:
I've now added fixes for all warnings except:

1. exported method ImportString returns unexported type *jsonnet.valueString, which can be annoying to use
1. exported method ImportCode returns unexported type jsonnet.value, which can be annoying to use
1. should have a package comment, unless it's in another file for this package

The first two would entail a much larger impact to the code, so I let them slide.

I tried to resolve the third warning by adding a package comment to vm.go. However, that only fixed it for the one file, since golint doesn't do global analysis. https://goo.gl/78pF4a

ast/location.go
```diff
@@ -93,22 +95,30 @@ func (lr *LocationRange) String() string {
 	return fmt.Sprintf("%s(%v)-(%v)", filePrefix, lr.Begin.String(), lr.End.String())
 }
 
-func (l *LocationRange) WithCode() bool {
-	return l.Begin.Line != 0
+// WithCode returns true iff the LocationRange is linked to code.
+// TODO: This is identical to lr.IsSet(). Is it required at all?
```
marcelocantos: I'm happy to fix, revert or raise an issue here.

ast/location.go
```diff
@@ -126,6 +136,8 @@ func (sp *SourceProvider) GetSnippet(loc LocationRange) string {
 	return result.String()
 }
 
+// BuildSource transforms a source file string into a Source struct.
+// TODO: This seems like a job for strings.Split() with a final \n touch-up.
```
marcelocantos: Ditto

error_formatter.go
```diff
@@ -36,8 +37,10 @@ type ErrorFormatter interface {
 	SetColorFormatter(color ColorFormatter)
 }
 
+// ColorFormatter represents a function that writes to the terminal using color.
 type ColorFormatter func(w io.Writer, f string, a ...interface{}) (n int, err error)
 
+// TODO: Explain why this is needed.
 var _ ErrorFormatter = &termErrorFormatter{}
```
marcelocantos: I was tempted to change this to `var _ = &termErrorFormatter{}`. The type seems redundant.

imports.go
```diff
@@ -140,18 +149,20 @@ func (importer *FileImporter) Import(dir, importedPath string) (*ImportedData, e
 	}
 
 	if !found {
-		return nil, fmt.Errorf("Couldn't open import %#v: No match locally or in the Jsonnet library paths.", importedPath)
+		return nil, fmt.Errorf("couldn't open import %#v: no match locally or in the Jsonnet library paths", importedPath)
```
marcelocantos: I wasn't sure this was a good idea. I was worried I might be breaking parity with C++ Jsonnet. The other issue that many other error messages that don't appear as `fmt.Errorf` parameters weren't picked up.

Ditto below.

linter/find_variables.go
```diff
@@ -5,10 +5,10 @@ import (
 	"github.com/google/go-jsonnet/parser"
 )
 
-type vScope map[ast.Identifier]*Variable
+type vScope map[ast.Identifier]*variable
```
marcelocantos: Variable wasn't used outside the package so I changed it to a private symbol. If the intent is to make it public somewhere down the line, I can add doc-comments instead.

parser/lexer.go
marcelocantos: I left the warning in place for a couple of other cases. Happy to revert if you prefer.

marcelocantos:
I noticed that doc.go has a valid package comment, so I removed the one I added to vm.go.

ast/location.go
sbarzowski: I think you're right. Feel free to do it in this change or a separate one later, whatever is more convenient.

ast/location.go
```diff
@@ -176,7 +188,7 @@ func LineBeginning(loc *LocationRange) LocationRange {
 	}
 }
 
-// lineEnding returns a part of the line directly after LocationRange
+// LineEnding returns a part of the line directly after LocationRange
```
sbarzowski: Same as above, "the part of a line" would be better.

error_formatter.go
sbarzowski: This is a Go idiom for ensuring that a struct implements an interface. 
More info: https://golang.org/doc/effective_go.html#blank_implements

imports.go
sbarzowski: FYI we don't have error message parity, no need to worry about it. Error messages are not part of the formal spec and may depend on implementation details. So we are not constrained here.

linter/find_variables.go
sbarzowski: Let's keep it private for now, that's ok.

(I think I had some plans for it being used in the future documentation generator or something, but I'm not sure, no point in keeping stuff public "just in case".)

thunks.go
```diff
@@ -181,6 +181,8 @@ func (f *bindingsUnboundField) bindToObject(sb selfBinding, origBindings binding
 	return f.inner.bindToObject(sb, upValues, fieldName)
 }
 
+// PlusSuperUnboundField represents a `field+: ...` that hasn't been bound to a
+// super object.
```
sbarzowski: `s/a super object/an object/`

It's not bound to any object. "not bound to a super object" suggests that it may be bound to an object in which it appeared.

parser/lexer.go
sbarzowski: It's ok.

parser/context.go
```diff
@@ -278,6 +278,7 @@ func specialChildren(node ast.Node) []ast.Node {
 	panic(fmt.Sprintf("specialChildren: Unknown node %#v", node))
 }
 
+// Children returns the direct, thunk and special children of a node.
```
sbarzowski: Nit: It may be better to just say it returns _all_ children. 

Direct, thunk and special children are slightly weird, unexported concepts.

imports.go
sbarzowski: For now it is ok, there is no explicit convention yet. Once we decide on some convention, we'll need to document it and keep it consistent everywhere.

sbarzowski:
This is a great step forward in cleaning up the code. I pointed out some minor oversights (unavoidable with a change that wide).

I'll look into the remaining golint warning after this is merged.

parser/lexer.go
sparkprime: What are the implications of `Tokens` being public but `token` being private?  Does it mean all you can do with Tokens is to give it to the parser?

marcelocantos:
I've applied all the suggested changes.

marcelocantos:
I just realised that I neglected to golint the full tree (🤦‍♂️). Shall I leave this one as is and start another PR to keep things contained?

parser/lexer.go
sbarzowski: Yeah, I think so. Tokens is probably completely opaque outside this package.

sbarzowski:
This change is very good as it is. It will certainly be a bit easier to review if further changes went to a new PR, but please do whatever is easier for you.

marcelocantos:
A separate PR is easy enough. I'll wait till this lands in master.

sbarzowski:
LGTM @sparkprime

marcelocantos:
I have the final round of fixes ready to go. The only remaining items are:

```bash
$ golint -min_confidence 0 ./...
imports.go:81:85: exported method ImportString returns unexported type *jsonnet.valueString, which can be annoying to use
imports.go:103:83: exported method ImportCode returns unexported type jsonnet.value, which can be annoying to use
thunks.go:299:80: exported method EvalCall returns unexported type jsonnet.value, which can be annoying to use
ast/ast.go:382:2: struct field Id should be ID
ast/ast.go:507:2: struct field Id should be ID
ast/ast.go:588:2: struct field Id should be ID
ast/ast.go:644:2: struct field Id should be ID
```

Since this PR is still awaiting a merge, I'm happy to apply the latest changes here or in a follow-up PR. Which do you prefer?

sparkprime:
I merged as this one was already big enough :)  Thanks for your work improving the health of the codebase!

Issue #200, opened 2018-02-26
Test cmdline handling edge cases


marcelocantos:
While working on #380, I accidentally broke the command-line processing logic because golint produced a spurious recommendation:

![image](https://user-images.githubusercontent.com/215143/36689254-fd9c45b6-1b82-11e8-8887-31883867ad9a.png)

```bash
$ golint -min_confidence 0 ./jsonnet
jsonnet/cmd.go:324:11: if block ends with a return statement, so drop this else and outdent its block
```

Automated tests didn't pick this up.

Need more cmdline tests.

marcelocantos:
https://github.com/golang/lint/issues/382

sbarzowski:
Just to be clear, have you run "./tests.sh"? `go test -v ./...` doesn't run tests from the C++ test suite.

BTW This seems to be the right place to add commandline tests: https://github.com/google/jsonnet/blob/master/test_cmd/run_cmd_tests.sh

BTW2 Something like `jsonnet -S -S -S -S -e '"blah"'` should catch it, right?

marcelocantos:
I tried it just now and I can't figure it out:

```bash
$ ./tests.sh
./tests.sh: line 8: tests_path.source: No such file or directory
```

It wants a `tests.source`, presumably to set the `TESTS_PATH`, which it will then cd to and run another `tests.sh`. But there's no other `tests.sh` in the code base.

sbarzowski:
Ah, right, it's described nowhere in the README and it's a tad bit hacky.

You need to create `tests.source` containing `export TESTS_PATH="/path/to/cpp-jsonnet/`

sparkprime:
We could also git submodule cpp jsonnet into go-jsonnet, but people seem to hate that

marcelocantos:
I don't hate submodules. It would be much cleaner from a testing perspective, and could even be added to CI. It would be super easy to add a line to tests.sh to ensure the submodule is up to date before running the tests, thus almost no one would need to deal with it explicitly.

marcelocantos:
#202 adds the google/jsonnet as a submodule.

marcelocantos:
Assuming the C++ tests cover enough cmdline cases, I guess not. I confirmed that the problem that triggered this issue (golint recommending spurious else-removal) is covered.

PR #201, opened 2018-02-26
Apply more golint recommendations

marcelocantos:
… part 2, continuing on from #196


marcelocantos:
I found that golint sometimes recommends structural changes that change the logic (see #200 and https://github.com/golang/lint/issues/382).

I made sure I didn't apply the spurious recommendations, but you should check any modified if-else-return combos especially carefully just to be sure.

ast/stdast.go
```diff
@@ -2768,6 +2768,8 @@ var p1 = &Source{
 		"\n",
 	},
 }
+
+// StdAst is the AST for the standard library.
```
sbarzowski: This file is autogenerated in `cmd/dumpstdlibast.go` and package `dump`. This comment needs to be generated, rather than inserted directly. An easy way to do it is to add `VariableConfig string`  to `dump.Options` and update `dump.dumpState.dump(...)` to use that comment.

sbarzowski: Instructions to regenerate this file may be handy: https://github.com/google/go-jsonnet#generated-stdlib

dump/dump.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package dump provides functions to dump values to stdout and strings.
```
sbarzowski: Perhaps you can take the first sentence from this package's readme, i.e:
Package dump can dump a Go data structure to Go source file, so that it can be statically embedded into other code.

ast/stdast.go
marcelocantos: Done

ast/stdast.go
```diff
@@ -1,3 +1,10 @@
+///////////////////////////////////////////////////////////
+// This file was auto-generated by cmd/dumpstdlibast.go. //
+// https://github.com/google/go-jsonnet#generated-stdlib //
+//                                                       //
+// --------------- DO NOT EDIT BY HAND! ---------------  //
```
sbarzowski: Nice!

sbarzowski:
LGTM @sparkprime

sparkprime:
thanks again!

PR #202, opened 2018-03-02
Add C++-Jsonnet submodule for tests

marcelocantos:
Also:

- Run `tests.sh` in travis-ci.
- `tests.sh` updates the submodule and run its tests against go-jsonnet.
- `tests.sh` also runs `go test ./...` (but not in travis-ci, which does so via goverage).
- Update `README.md` to explain how to run tests.

Currently, the submodule points to the latest Jsonnet release, v0.9.5.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cpp-jsonnet"]
+	path = cpp-jsonnet
+	url = https://github.com/google/jsonnet.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,4 +10,4 @@ before_install:
   - if ! go get github.com/golang/tools/cmd/cover; then go get golang.org/x/tools/cmd/cover; fi
 script:
   - $HOME/gopath/bin/goveralls -service=travis-ci
-  - ./tests.sh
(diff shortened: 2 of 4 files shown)


sbarzowski:
Could you add the instructions for running tests to the README? (`go test ./... && ./tests.sh`). Or maybe we should add `go test ./...` part to `tests.sh` now, so that there's one script which does the whole thing.

marcelocantos:
@sbarzowski Happy to, but I'm wondering whether goveralls does some or all of that already in the current .travis.yml.

sbarzowski:
Yeah, it does. I wanted a single script to run tests on dev machine. This change is not about that, so let's leave it for now, unless you have a good solution in mind already.

marcelocantos:
I added it anyway. `go test ./...` runs very quickly, so it won't kill travis to run it twice.

marcelocantos:
I added a parameter to tests.sh to allow travis to skip `go test ./...`.

sbarzowski:
LGTM @sparkprime

sparkprime:
This will definitely help avoid the drift

PR #210, opened 2018-03-07
Clone ASTs to avoid aliasing and double-unescaping

sparkprime:
@sbarzowski @marcelocantos @benley

### DIFF
--- a/ast/clone.go
+++ b/ast/clone.go
@@ -290,7 +290,7 @@ func clone(astPtr *Node) {
 		*r = *node
 
 	default:
-		panic(fmt.Sprintf("Desugarer does not recognize ast: %s", reflect.TypeOf(node)))
+		panic(fmt.Sprintf("ast.Clone() does not recognize ast: %s", reflect.TypeOf(node)))
 	}
 
 	cloneNodeBase(*astPtr)
--- a/ast/clone.go
+++ b/ast/clone.go
@@ -1,0 +1,302 @@
+/*
+Copyright 2018 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


ast/clone.go
```diff
@@ -0,0 +1,302 @@
...
+func cloneNodeBase(astPtr Node) {
+	if astPtr.Context() != nil {
+		newContext := new(string)
+		*newContext = *astPtr.Context()
+		astPtr.SetContext(newContext)
+	}
+	astPtr.SetFreeVariables(append(make(Identifiers, 0), astPtr.FreeVariables()...))
+}
+
+// Updates *astPtr to point to a deep clone of what it originally pointed at.
```
sbarzowski: "In-place clone" is a bit weird. I see it saves quite a lot of boilerplate, so it's ok. I'm worried it could be a bit bugprone - it's easy to forget to clone a field. Checking that the result is a proper tree would help.

The check could be something along these lines:
```
func check(root *Node) {
    visited := make(map [Node*]bool);
    func dfs(node *Node) {
          visited[node] = true;
          for child := range children(node) {
              if (visited[child]) {
                    error("Not a tree");
                    return;
              }
              dfs(node);
          }
          dfs(root);
    }
```

ast/clone.go
```diff
@@ -0,0 +1,302 @@
...
+		*r = *node
+		clone(&r.Expr)
+
+	case *Var:
+		r := new(Var)
+		*astPtr = r
+		*r = *node
+
+	default:
+		panic(fmt.Sprintf("Desugarer does not recognize ast: %s", reflect.TypeOf(node)))
```
sbarzowski: s/Desugarer/Clone/ ?

ast/clone.go
sparkprime: Yeah, although even the most explicit option: `&Foo { x: other.x, y: clone(other.y), ... }` is liable to have you forget to clone stuff because you didn't realise it's a pointer (easy given the typedefs, interfaces). I think it's OK because we'll want that checker anyway.  It's tricky to give a good error message pointing to where in the tree the duplication occurred, though!

sparkprime: I guess the path of types (concrete types, not Node) is easy enough to provide (just need an explicit stack of Node and use reflection) and specific enough to find the problem.
visited needs to map Node to the path though, so we can say where it was found elsewhere.

sbarzowski: Perhaps it's possible to record the path or reuse the context for that.

It's only an assertion for tests, not a user-facing message, so I wouldn't spend too much time on making it better.

PR #219, opened 2018-04-12
godoc.org-compatible ast package

rogpeppe:
Change the dump code so that it hides the values of variable
definitions if they're large. This means that godoc.org should
be able to deal with the output, and the godoc output is
readable without needing to read through a huge struct literal definition.

Other approaches might be to always generate an extra variable
(seems unnecessary) or to pass to writer explicitly to the dump
methods rather than swapping s.w out temporarily. The former
seems unnecessarily intrusive to the usual output; the latter
seemed unnecessarily intrusive to the source itself.
YMMV.

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -184,7 +184,21 @@ func (s *dumpState) dump(value interface{}) {
 	if v.Kind() == reflect.Ptr && v.IsNil() {
 		printNil(s.w)
 	} else {
+		// Replace the writer with a temporary buffer so that
+		// we can see how big the output is. If it's too big,
+		// we use an intermediate value so that godoc output
+		// isn't swamped by the literal.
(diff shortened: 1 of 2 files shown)


sbarzowski:
Thanks for the contribution!

Sounds good. In fact, ideally stdast.go wouldn't appear in the docs at all. It's just a hairy, auto-generated implementation detail (pre-parsed standard library).

@sparkprime

rogpeppe:
@sparkprime As long as ast types are mentioned in the core jsonnet godoc, I think it's worth making the ast package godoc-compatible. It might be worth hiding most of that package inside an internal package and exposing only some chosen parts of it, designed for external consumption. For the time being though, I think it's worth doing something like this to make it possible to see what can be done with ast.Node and friends.

PR #221, opened 2018-05-09
Implement operators == and != as builtins

sbarzowski:
People use these operators in tight loops, without even
thinking about it. It's previous implementation required
multiple object lookups (std.), string comparisons (for types)
and multiple jsonnet function calls.

This change introduces a builtin, efficient implementation.
It results in ~3x speedup in strContains benchmark that
Angus provided on Slack.

Additional benefit is that primitiveEquals is now obsolete
and simplification is always good.


sbarzowski:
@sparkprime @anguslees

sbarzowski:
This failed test is actually an improvement. In C++ test suite there is a test which checks that Go implementation goes over the stack limit. We have only one function call to evaluate `==`, so the stack size is reduced.

I'll send a separate PR to cpp-jsonnet later.

sparkprime:
I think my changes would add to the performance improvement here.

One of the things I did is change all the builtins to take raw values instead of potentialValue

sparkprime:
I tried expanding the test case by 10x but got hit by the lack of tail call optimization in Go.  We're going to have to do something about that eventually.  (Hopefully the solution won't kill performance).

sparkprime:
Also it'd be good for that max_stack4 test to be more robust to the implementation of ==, although I'm not sure how.  It's possible that the goal of "having the two implementations be consistent in how many stack frames are necessary to execute a given piece of Jsonnet code" is actually not achievable.

sbarzowski:
> It's possible that the goal of "having the two implementations be consistent in how many stack frames are necessary to execute a given piece of Jsonnet code" is actually not achievable.

It's hard to imagine it staying consistent even across minor versions of the same implementation. It's an implementation detail. And I don't really see the point of trying to keep it the same. Isn't it just a rough check for endless recursion?

sparkprime:
I rebased this on top of my branch:

```
dcunnin@dcunnin7:~$ time go-jsonnet str_contains.jsonnet
RUNTIME ERROR: Assertion failed

        During evaluation


real    0m1.272s
user    0m1.584s
sys     0m0.316s
dcunnin@dcunnin7:~$ time jsonnet str_contains.jsonnet
RUNTIME ERROR: Assertion failed.
        str_contains.jsonnet:(17:1)-(19:5)

real    0m1.132s
user    0m1.116s
sys     0m0.012s
```

sparkprime:
It's an implementation detail of the libraries, yes.  It's a bit surprising that == is a library call though.

More importantly the concept of a stackframe is not specified.

PR #231, opened 2018-06-11
Lex fodder in a manner compatible with C++

sparkprime:
I also rewrote lexer_test to use individual tests for each case instead of iterating through an array.  While not as dense on the page, the density of information is about the same and this way you can get more useful stacktraces as well as easily select specific tests to run.

### DIFF
--- a/ast/fodder.go
+++ b/ast/fodder.go
@@ -1,0 +1,179 @@
+/*
+Copyright 2018 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package ast provides AST nodes and ancillary structures and algorithms.
+package ast
+
+import (
+	"fmt"
+)
+
+// Fodder
+
+// FodderKind is an enum.
+type FodderKind int
+
+const (
+	// FodderLineEnd represents a line ending.
+	//
+	// It indicates that the next token, paragraph, or interstitial
+	// should be on a new line.
+	//
+	// A single comment string is allowed, which flows before the new line.
+	//
+	// The LineEnd fodder specifies the indentation level and vertical spacing
+	// before whatever comes next.
+	FodderLineEnd FodderKind = iota
+
+	// FodderInterstitial represents a comment in middle of a line.
+	//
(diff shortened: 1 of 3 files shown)


sparkprime:
The tests didn't build because of a stray file (not in the PR) and looked like they succeeded but actually there is quite a lot more work for me to do here

sparkprime:
Ok it actually wasn't that hard.

ast/fodder.go
```diff
@@ -0,0 +1,179 @@
...
+	// interstitial, it will also flow after it on the same line.  If it follows
+	// a new line or a paragraph, it is the first thing on the following line,
+	// after the blank lines and indentation specified by the previous fodder.
+	//
+	// There is exactly one comment string.
+	FodderInterstitial
+
+	// FodderParagraph represents a comment consisting of at least one line.
+	//
+	// // and # style commes have exactly one line.  C-style comments can have
```
sbarzowski: commes -> comments

parser/lexer.go
```diff
@@ -209,6 +188,47 @@ func isSymbol(r rune) bool {
 	return false
 }
 
+func isHorzWs(r rune) bool {
```
sbarzowski: nit: I wouldn't abbreviate it. I would go with`isHorizontalWhitespace`.

sbarzowski: Same with other functions with "Ws" in their names.

parser/lexer_test.go
```diff
@@ -272,24 +193,628 @@ func tokensEqual(ts1, ts2 Tokens) bool {
...
+			errString, expectedError)
+	}
+	if err == nil && !tokensEqual(tokens, testTokens) {
+		t.Errorf("got\n\t%+v\nexpected\n\t%+v", tokens, expected)
+	}
+}
+
+// TODO: test position reporting
+
+func TestEmpty(t *testing.T) {
```
sbarzowski: What was wrong with the old table-style tests?

sparkprime: Impossible to run a single test in isolation without modifying the code.  Although Go doesn't seem to do this stuff yet, other frameworks prefer breaking stuff up as much as possible to enable individual running of tests, reporting of errors at finer granularity, running tests in parallel, and helping avoid state accumulating between tests.

ast/fodder.go
sparkprime: Done.

parser/lexer.go
sparkprime: Done.

parser/lexer_test.go
sbarzowski: I'm ok with that.

Just FYI the way it worked before was **not** the best you could get from Table Driven Tests. All of the stuff you mentioned can be achieved with TDT. Most importantly it lacked `t.Run(tt.in, func(t *testing.T) {...` - so all the cases were treated as one test. If we added that one line with `t.Run(...` each testcase would be a subtest and we could easily run them separately and in parallel.
More info: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

PR #232, opened 2018-06-11
pull Local.Binds[].Body out via specialChildren

jamesonjlee:
this allows traversing the Binds in a Local via the parser.Children() call on a Local Node.

### DIFF
--- a/parser/context.go
+++ b/parser/context.go
@@ -249,7 +249,14 @@ func specialChildren(node ast.Node) []ast.Node {
 	case *ast.Slice:
 		return nil
 	case *ast.Local:
-		return nil
+		if len(node.Binds) == 0 {
+			return nil
+		}
+		children := make([]ast.Node, 0, len(node.Binds))
+		for _, bind := range node.Binds {
+			children = append(children, bind.Body)
+		}
+		return children
 	case *ast.LiteralBoolean:
 		return nil
 	case *ast.LiteralNull:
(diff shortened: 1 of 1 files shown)


parser/context.go
```diff
@@ -249,7 +249,14 @@ func specialChildren(node ast.Node) []ast.Node {
...
 	case *ast.Local:
-		return nil
+		if len(node.Binds) == 0 {
+			return nil
+		}
+		children := make([]ast.Node, 0, len(node.Binds))
+		for _, bind := range node.Binds {
+			children = append(children, bind.Body)
+		}
+		return children
```
sbarzowski: Actually local body should also be "special" rather than "direct". Could you add it to the children here?

parser/context.go
```diff
@@ -249,7 +249,14 @@ func specialChildren(node ast.Node) []ast.Node {
 	case *ast.Slice:
 		return nil
 	case *ast.Local:
-		return nil
+		if len(node.Binds) == 0 {
```
sbarzowski: Is it necessary? Especially if we add `body` (see comment below), but generally it seems that:
a) That's impossible - at least one binding is required 
b) I think we could just make a 0-length slice using the code below

parser/context.go
jamesonjlee: done

parser/context.go
jamesonjlee: done

sbarzowski:
Awesome, thanks.

LGTM @sparkprime

PR #234, opened 2018-06-22
Adds TravisCI release process

lwander:
Given the <tag> chosen in the GitHub releases page, this builds jsonnet
for darwin & linux amd64 (we can add more if needed), and uploads to
`gs://jsonnet/<tag>/<os>/<arch>/jsonnet`


lwander:
It looks like I'm unable to encrypt the json key using the travis gem... maybe a lack of permissions in this repo? @sparkprime

lwander:
Turns out the key can only be decrypted during during a release -- should be good to go now.

sparkprime:
Should we kick it off for the last few previous releases somehow?

lwander:
I think GH releases allows you to re-tag prior releases. If that's possible, I would run it.

PR #242, opened 2018-08-19
Fix build instructions

garethr:
Without downloading the dependency the current instructions fail to build.


sparkprime:
Gonna merge it, what sbarzowski proposed can be a future change

PR #243, opened 2018-08-29
added native exec support

sourcec0de:
What do you think about adding support for a native exec command build in?

- allows jsonnet to exec an arbitrary command
- if that command returns valid json it can be used
- if it fails to return valid json the exec command fails it errors

```jsonnet
local terraform_output = std.native('exec')('terraform', ['output', '-json'])
```

### DIFF
--- a/native_exec.go
+++ b/native_exec.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jsonnet
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os/exec"
+	"reflect"
+	"strings"
+
+	"github.com/google/go-jsonnet/ast"
+)
+
+func execCommand(cmd *exec.Cmd) (*bytes.Buffer, error) {
+	stderr := &bytes.Buffer{}
+	stdout := &bytes.Buffer{}
+	cmd.Stderr = stderr
+	cmd.Stdout = stdout
+	if err := cmd.Run(); err != nil {
+		cmdStr := cmd.Path + " " + strings.Join(cmd.Args, " ")
+		if stderr.Len() == 0 {
+			return stdout, fmt.Errorf("%s: %v", cmdStr, err)
+		}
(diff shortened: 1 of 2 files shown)


sparkprime:
It seems that in this form it would be quite hard to write configs that behave the same on different systems because the executed binaries might be slightly different (e.g. different Linux distros), and that's assuming that everything is properly hermetic, which it probably won't be.

You can do `jsonnet --ext-code foo=$(my command)` or use temporary files.  The advantage of temporary files is that they provide an audit trail if you need to find out why the config did what it did, and potentially even the ability to replay the config execution without re-executing the command that generated the temporary files.

vitiral:
So there is another use case I want to keep in mind: using jsonnet (and native functions) as part of a deterministic build system, something like Nix or Bazel. In such a case the "native functions" (binaries) themselves could be made to be deterministic, and while native functions could create side effects, it is the job of the build system to not allow them.

On your proposal specifically, I had a similar proposal [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jsonnet/OhWV2RKlUFE/3EHMphMSBgAJ). My rough proposed API would be:

```
std.extern_function(exc, args=[], stdin={}, env=std.ENV)
```

And here is an example use:
```
from_yaml(path) =
    local from_yaml_py = |||
        import sys, json, yaml
        args = json.load(sys.stdin)
        with open(args['path']) as f:
            value = yaml.load(f)
        sys.stdout.write(json.dump(value))

    std.extern_function("python", ["-c", from_yaml_py], {path: path})
```

Just like your proposal it would pass large items though stdin (serialized as json) and receive the results via stdout.

Unlike your proposal it has `args` and `env` allowing the control of the environment and would not worry about the binaries being on any specific directory (just have to be on PATH). I can see why there would be arguments against that, but I would strongly caution against trying to use a "single directory" because in a build system use case there may be multiple binaries of the same name (although the build system could just slap a hash on that name if it wanted... so maybe it doesn't matter).

I think being able to define these functions **inline** within scripting-like languages like python, bash and even Go would bring a lot of power to jsonnet. Power which must be used wisely, but would be extremely useful.

vitiral:
One major difference from your proposal: I want the output to be able to itself be **jsonnet**. This could allow some very interesting use cases, like maybe allowing other configuration languages to export jsonnet functions, etc.

sparkprime:
Does it have to be inline or could it be provided on the commandline somehow?  At which point you may as well refer to a built executable (or a shell script on disk).  It could be any language.

I think this is a bit more likely to be hermetic because you'd have to specify exactly which executable you want, and it would be visible at the top level that the config has this dependency on this external bit of code.

vitiral:
I thought a little more about this use case over the weekend and am starting to feel the opposite of what I did before, at least for the "build system" use case. If this were used for the build system, then _that_ could perform whatever executions were necessary to "fill out" the rest of the tree -- jsonnet _itself_ doesn't need to be able to handle an external executable.

Not that there is _no_ use case for this... the primary one is probably being able to support arbitrary config languages (that convert to json), or being able to embed your own "jsonnet package manager" type script. However, I think the use case I put forward is actually an anti-use case.

Issue #237, opened 2018-06-30
Use vendoring


tamalsaha:
Have you considering adding vendoring? Otherwise provide pre-built binaries so we can just download and put in the bin folder?

seh:
For the record, I found that running _dep init_ in the root directory took care of vendoring the dependencies, per the following output:
```
  Locking in v0.0.9 (167de6b) for transitive dep github.com/mattn/go-colorable
  Using ^1.0.0 as constraint for direct dep github.com/sergi/go-diff
  Locking in v1.0.0 (1744e29) for direct dep github.com/sergi/go-diff
  Using ^1.7.0 as constraint for direct dep github.com/fatih/color
  Locking in v1.7.0 (5b77d2a) for direct dep github.com/fatih/color
  Locking in v0.0.4 (6ca4dbf) for transitive dep github.com/mattn/go-isatty
  Locking in master (66b7b13) for transitive dep golang.org/x/sys
```
Those were its default choices, which we could refine as necessary.

michaeljs1990:
I think this ticket can be closed out now with master using go modules unless we want to do more work around providing pre built binaries.

Issue #239, opened 2018-07-17
Fails to build


SleepyBrett:
```
 ~/Dropbox/go/src/github.com/google/go-jsonnet/jsonnet   master  go build
# github.com/google/go-jsonnet/jsonnet
./cmd.go:497:60: color.New(color.FgRed).Fprintf undefined (type *color.Color has no field or method Fprintf)

 ~/Dropbox/go/src/github.com/google/go-jsonnet/jsonnet   master  go version
go version go1.10.2 darwin/amd64
```

Some drift in the color library?

garethr:
I just ran into this and submitted a fix to the README: https://github.com/google/go-jsonnet/pull/242

sbarzowski:
Thanks for pointing this out.

The reason it happens is that only the CLI depends on the color package. Formally CLI is a separate package from `github.com/google/go-jsonnet`. Jsonnet library doesn't depend on the CLI (only the other way around), so `go get github.com/google/go-jsonnet` doesn't install dependencies of the CLI.

SleepyBrett:
ah gotcha!

mmolnar:
Sorry but this still does not work for me, library `github.com/fatih/color` really does not have method Fprintf so `color.New(color.FgRed).Fprintf` will throw an error even if you run go get.

sbarzowski:
I think it does - https://github.com/fatih/color/blob/master/color.go#L214 and it worked for us for a while. Any chance you have some very ancient version of `github.com/fatih/color` and you could update? If that's not it, please paste the details (exact versions, and error message), so we can debug - since we have "works on my machine" kind of situation.

mmolnar:
That was my error, you are right, sorry for bothering you.

PR #244, opened 2018-09-18
vendor: Add vendoring with dep

dlespiau:
Vendoring helps with reproducible builds and make compiling go-jsonnet easier
as it doesn't rely on having its dependencies in GOPATH.

dep is the current vendoring tool with the most mind share and this commit is
just the result of:

  $ dep init

Fixes: #237


dlespiau:
Something seems wrong with he Appveyor test. It also fails on master.

```
build started
git clone -q https://github.com/google/go-jsonnet.git C:\projects\go-jsonnet
git fetch -q origin +refs/pull/244/merge:
git checkout -qf FETCH_HEAD
The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.
```

sbarzowski:
Thanks! This would definitely help with the problems people experienced on #239.

I'm not sure how vendor dir would affect people using go-jsonnet as a library, especially if they don't use dep, but happen to depend on the same packages. Do you have any experience with that? We need to have an answer ready.

And eventually I think we should use go modules - at least once they stabilize and become the fully official solution.

Don't worry about appveyor. It was set up once, and caused problems ever since. We probably should either fix it or remove it, so it stops bothering people.

dlespiau:
Yes, indeed! It should help in that case. Next PR could be simplifying the install instructions to a single `go get` command (or it could be a commit on top of this PR).
 
For library users:
- if they use dep, nice! dep should solve the constraints various libraries have and come up with the right version to use.
- if they don't, it's always possible to `rm -rf vendor/`. I would expect other vendoring tools to not look at the `vendor/` directory of dependencies if they haven't been vendored using that same tool, but it has been a very long time since I looked at anything else than `dep`. A nice plus is that, when vendoring their own project with other tools, people can look at the `Gopkg.lock` file to have guidance on the versions that go-jsonnet was tested with.

There's another option: it's possible to commit the Gopkg.toml and Gopkg.lock files and **not** commit the vendor directory. It means people have to run `dep ensure` themselves before building jsonnet, dep ensure will then select the "locked" versions described in the `.lock` file. `go get` will not work though, but I understand some maintainers may prefer not having the `vendor/` directory committed.

Everyone should move to go modules once they are ready :)

dlespiau:
Added a couple of commit on top that result from vendoring dependencies:
- We don't need to `go get github.com/sergi/go-diff/diffmatchpatch` in the CI setup as it's vendored
- We can simplify the installation instructions in the README

michaeljs1990:
Does it make sense to use dep when `go mod` is becoming the lingua franca in the golang ecosystem since 1.11?

dlespiau:
No, not really, time to close this PR!

PR #252, opened 2018-12-04
Automatically create output folders

lorenz:
This is an implementation of https://github.com/google/jsonnet/issues/195 for the Go version since the concerns mentioned in that issue are of no concern in a Go application. It's just a very quick fix done for my own needs, feel free to reject in case you don't want any divergence from the C++ implementation.

### DIFF
--- a/jsonnet/cmd.go
+++ b/jsonnet/cmd.go
@@ -86,6 +86,7 @@ func usage(o io.Writer) {
 	fmt.Fprintln(o, "  -J / --jpath <dir>      Specify an additional library search dir")
 	fmt.Fprintln(o, "  -o / --output-file <file> Write to the output file rather than stdout")
 	fmt.Fprintln(o, "  -m / --multi <dir>      Write multiple files to the directory, list files on stdout")
+	fmt.Fprintln(o, "  -c / --create-output-dirs  Automatically creates all parent directories for files")
 	fmt.Fprintln(o, "  -y / --yaml-stream      Write output as a YAML stream of JSON documents")
 	fmt.Fprintln(o, "  -S / --string           Expect a string, manifest as plain text")
 	fmt.Fprintln(o, "  -s / --max-stack <n>    Number of allowed stack frames")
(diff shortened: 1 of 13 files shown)


lorenz:
Signed the CLA

sparkprime:
This is great but can you put it in a flag that's off by default and also make it do it in the single file case?

lorenz:
Sure

lorenz:
Made it into an option and added single file support

lorenz:
The failed tests are in the cpp-jsonnet repo, how should this be synchronized? If the PR there goes in before this one, the build is broken and the other way round too.

sbarzowski:
Oh, we're comparing help output directly. Well this is our test mechanism's fault, sorry about that. It's not very important how we go around it, as long as we update both in the timespan of less than few hours. Pehaps later we can move implementation-specific output to implementation repo (it should be just a few lines of changes in the scripts).

sparkprime:
I believe it should be possible to update cpp-jsonnet but not go-jsonnet, and then go-jsonnet will not take that version of cpp-jsonnet until it is updated, so you can do that update at the same time as introducing the change.

lorenz:
Update seems to have broken another assertion (non cmd-related), not sure about that

sbarzowski:
I don't think it has anything to do with this change. It's happened just because of the submodule update. I'll investigate later today.

sbarzowski:
As I thought, it's just the matter of updating stdlib to match C++ updates (there was a bugfix related to escaping there). If you add this commit to the pull request the tests should run fine https://github.com/sbarzowski/go-jsonnet/commit/baff839b5c47ce94c99bc98cc591fb1bd57b3bc5

lorenz:
Looks good now

sbarzowski:
Yes, I'm officially ok with my commit being merged here.

sparkprime:
Thanks

alem0lars:
Is this feature released?

sbarzowski:
Yes

PR #257, opened 2019-02-01
Minimal C bindings

sbarzowski:


sbarzowski:
This is a continuation of my work from summer 2017. In the final version of this PR I intend to make the basic functionality available, but skip the tricky parts (e.g. support for user provided native functions). 

Current WIP version is very messy and allows only creating a VM, setting max stack size, setting string vs json output, evaluating a snippet and destroying a VM.

It's possible to play with it by running `run.sh` from within `compat_test` directory. Note that you actually need to be in that directory for it to work (it uses relative paths). It compiles the C bindings and runs `compat_test/compat_test.py` using them.

sbarzowski:
Added support for jpath_add and max_trace

sbarzowski:
Added support for jsonnet_evaluate_file

I think now I will try to clean it up a little and make sure that memory allocation business is correctly handled (I forgot why we needed the whole jsonnet_realloc thing). 

I suppose we can skip multi and stream variants in this PR.

sparkprime:
So the issue is that the client is responsible for memory management in that API.  You can't make that a no-op because then you're relying on Go cleaning it up, which it can't do because it does not know if anyone on the C side is still using it.  So, instead you have to alloc those buffers, copy to them, and hand off the pointers to the C client, trusting it to eventually call realloc to clean it up.  I don't know what's available in Go in this area.  I'm happy to chat about it on Slack though and we can figure it out.

sbarzowski:
@sparkprime 
I think it should has reached a usable state. Please take a look.

Regarding memory it was quite easy, unless I missed something. A regular malloc/free are used under the hood. Some of the allocations happen through cgo are calling malloc (that's what `C.CString` does). So `jsonnet_realloc` just runs standard `malloc`/`realloc`/`free` depending on the situation and crashes on failure.

The tests for the bindings are now run as part of `tests.sh` (so it is going to be checked by Travis).

sparkprime:
I had a play with it (haven't read the code yet).  I built your .so and then build the jsonnet C++ commandline against the C++ .so (usually it statically links it).  That let me easily switch the C++ CLI tool between C++ and Go .so.  So e.g. you get a --help talking about `fmt`, but you get the error:

```
dcunnin@dcunnin1:~/go-jsonnet/compat$ jsonnet_dyn
ERROR: must give filename

Jsonnet commandline interpreter go-jsonnet

General commandline:
jsonnet [<cmd>] {<option>} { <filename> }
Note: <cmd> defaults to "eval"

The eval command:
jsonnet eval {<option>} <filename>
Note: Only one filename is supported

Available eval options:
  -h / --help             This message
  -e / --exec             Treat filename as code
  -J / --jpath <dir>      Specify an additional library search dir (right-most wins)
  -o / --output-file <file> Write to the output file rather than stdout
  -m / --multi <dir>      Write multiple files to the directory, list files on stdout
  -y / --yaml-stream      Write output as a YAML stream of JSON documents
  -S / --string           Expect a string, manifest as plain text
  -s / --max-stack <n>    Number of allowed stack frames
  -t / --max-trace <n>    Max length of stack trace before cropping
  --gc-min-objects <n>    Do not run garbage collector until this many
  --gc-growth-trigger <n> Run garbage collector after this amount of object growth
  --version               Print version
Available options for specifying values of 'external' variables:
Provide the value as a string:
  -V / --ext-str <var>[=<val>]     If <val> is omitted, get from environment var <var>
       --ext-str-file <var>=<file> Read the string from the file
Provide a value as Jsonnet code:
  --ext-code <var>[=<code>]    If <code> is omitted, get from environment var <var>
  --ext-code-file <var>=<file> Read the code from the file
Available options for specifying values of 'top-level arguments':
Provide the value as a string:
  -A / --tla-str <var>[=<val>]     If <val> is omitted, get from environment var <var>
       --tla-str-file <var>=<file> Read the string from the file
Provide a value as Jsonnet code:
  --tla-code <var>[=<code>]    If <code> is omitted, get from environment var <var>
  --tla-code-file <var>=<file> Read the code from the file
Environment variables:
JSONNET_PATH is a colon (semicolon on Windows) separated list of directories added
in reverse order before the paths specified by --jpath (i.e. left-most wins)
E.g. JSONNET_PATH=a:b jsonnet -J c -J d is equivalent to:
JSONNET_PATH=d:c:a:b jsonnet
jsonnet -J b -J a -J c -J d

The fmt command:
jsonnet fmt {<option>} { <filename> }
Note: Some options do not support multiple filenames

Available fmt options:
  -h / --help             This message
  -e / --exec             Treat filename as code
  -o / --output-file <file> Write to the output file rather than stdout
  -i / --in-place         Update the Jsonnet file(s) in place.
  --test                  Exit with failure if reformatting changed the file(s).
  -n / --indent <n>       Number of spaces to indent by (default 2, 0 means no change)
  --max-blank-lines <n>   Max vertical spacing, 0 means no change (default 2)
  --string-style <d|s|l>  Enforce double, single (default) quotes or 'leave'
  --comment-style <h|s|l> # (h), // (s)(default), or 'leave'; never changes she-bang
  --[no-]pretty-field-names Use syntax sugar for fields and indexing (on by default)
  --[no-]pad-arrays       [ 1, 2, 3 ] instead of [1, 2, 3]
  --[no-]pad-objects      { x: 1, y: 2 } instead of {x: 1, y: 2} (on by default)
  --[no-]sort-imports     Sorting of imports (on by default)
  --debug-desugaring      Unparse the desugared AST without executing it
  --version               Print version

In all cases:
<filename> can be - (stdin)
Multichar options are expanded e.g. -abc becomes -a -b -c.
The -- option suppresses option processing for subsequent arguments.
Note that since filenames and jsonnet programs can begin with -, it is advised to
use -- if the argument is unknown, e.g. jsonnet -- "$FILENAME".
dcunnin@dcunnin1:~/go-jsonnet/compat$ jsonnet_dyn fmt -e '1'
FATAL ERROR: Not supported by Go implementation.
Aborted (core dumped)
```

That is working as intented

I also tried running the test suite using the CLI, which worked except for a change in the YAML pretty printing which isn't in your go-jsonnet branch yet (maybe not in go-jsonnet generally)

```
LD_LIBRARY_PATH=$HOME/go-jsonnet/compat JSONNET_BIN=jsonnet_dyn DISABLE_ERROR_TESTS=true DISABLE_FMT_TESTS=true ./run_tests.sh
```

and similarly for test_cmd:

```
LD_LIBRARY_PATH=$HOME/go-jsonnet/compat JSONNET_BIN=jsonnet_dyn DISABLE_ERROR_TESTS=true DISABLE_FMT_TESTS=true ./run_cmd_tests.sh
```

Those both worked, except for errors and the version number, i.e.:

```
dcunnin@dcunnin1:~/go-jsonnet/compat$ jsonnet_dyn --version
Jsonnet commandline interpreter go-jsonnet
```

Is not right

sparkprime:
Version comes from here at the moment https://github.com/sbarzowski/go-jsonnet/blob/master/compat/libjsonnet.h#L27

sbarzowski:
I changed the version string. It now returns "v0.12.0 (go-jsonnet)".

I think #260 should fix the YAML think, it's completely orthogonal to everything that's going on here.

compat/c_layer.go
```diff
@@ -1,48 +1,159 @@
...
-	out, err := vm.EvaluateSnippet("c", s)
-	if err != nil {
-		fmt.Println(err)
+// Because of Go GC, there are restrictions on keeping Go pointers in C.
+// We cannot just pass *jsonnet.VM to C. So instead we use "handle" structs in C
+// which refer to JsonnetVM by a numeric id.
+// The way it is implemented below is simple and has low overhead, but requires us to keep
+// a list of used IDs. This results in a permanent "leak". I don't expect it to ever
+// become a problem.
+var VMs = []*vm{}
```
sparkprime: Can you comment that 0 is never a valid id, and the index of VM in the array is always `id - 1`

compat/c_layer.go
```diff
@@ -1,48 +1,159 @@
...
-	fmt.Println(code)
-	s := C.GoString(code)
-	vm := jsonnet.MakeVM()
-	fmt.Println(s)
-	out, err := vm.EvaluateSnippet(f, s)
+//export jsonnet_destroy
+func jsonnet_destroy(vmRef *C.struct_JsonnetVm) {
+	VMs[vmRef.id-1] = nil
+	freedIDs = append(freedIDs, uint32(vmRef.id))
+	C.jsonnet_internal_free(vmRef)
```
sparkprime: I wonder if it's better to just declare `void free(void *ptr);` in internal.h?  Is there an explicit best practice for this?

sparkprime:
Curious where `#include "blah.h"` comes from in https://github.com/google/go-jsonnet/blob/master/compat/compat.h#L10

compat/c_layer.go
sparkprime: Looks like you can just do C.free(unsafe.Pointer(vmRef))

sparkprime: ...actually you're creating with `new` so you have to clean it up with `delete`, so you do need jsonnet_internal_free but it shouldn't call `free` in its implementation.

sparkprime:
I read all the code, generally looks good.  Some thoughts:
- The libjsonnet.h still has the fmt calls in it, but I moved them into their own header on the C++ side.  Can we get rid of them? I suspect it will mean the C++ jsonnet executable will no-longer link, but nobody wants that configuration anyway.  The Python module will still work because it never used *_fmt_*.
- Can we share libjsonnet.h with the C++ submodule?
- The native callback stuff is going to be non-trivial I guess because of the need to convert from the C++ json tree to the Go one and do all that memory management.

compat/c_layer.go
sbarzowski: done

compat/c_layer.go
sbarzowski: done

I think I could get rid of these functions by just using `malloc`/`free` on Go side, so I can also do that if that's better. Sizeof is a bit ugly on Go side, but it should work.

sbarzowski:
re "blah.h" - this whole file is unnecessary, I removed it

re removing fmt functions - done. I guess we could add an ifdef in cpp version if we really wanted to link with go version (it could also omit that part in help which would  be also good).

re sharing libjsonnet.h with C++ module - done

re native callback - I don't think it's very difficult - just a lot of stupid code which I don't think is necessary for this PR. When we're at it, shouldn't `json.cpp` include `json.h` with `extern "C"` - the way it is I think the names of `jsonnet_json_*`  functions are going to be mangled.

sbarzowski:
@sparkprime ^

sparkprime:
Woops, yes, nice job

PR #229, opened 2018-06-05
Add basic tests for Children and add missing stuff

sbarzowski:

PR #259, opened 2019-02-28
Smart strings

sbarzowski:
When adding long strings, don't copy them immediately. Instead
build long strings only when their contents are requested.

This allows to build a long string from parts, using a regular
operator+ in linear time. This lets users to worry much less
about using std.join etc.

If indexing the string is mixed with building it using operator+
the behavior can still be quadratic. We may want to address it in
a later change.

Depends on: #229

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -33,12 +33,12 @@ import (
 func builtinPlus(i *interpreter, trace traceElement, x, y value) (value, error) {
 	// TODO(sbarzowski) perhaps a more elegant way to dispatch
 	switch right := y.(type) {
-	case *valueString:
+	case valueString:
 		left, err := builtinToString(i, trace, x)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


sbarzowski:
Example benchmark:
```
std.foldl(function(e, res) e + res, std.makeArray(20000, function(i) "aaaaa"), "")
```

This version: 0.73s user 0.13s system 131% cpu 0.652 total
Current master: 1.64s user 1.89s system 106% cpu 3.306 total
C++ version: 1.24s user 0.61s system 98% cpu 1.872 total

sparkprime:
This is a good idea.  We should tune it on some realistic benchmarks if possible.

I think having a native deep join that will do 2 passes over the tree will still be faster in the case of manifest functions, but obviously that is more of an expert approach.

sbarzowski:
I rebased it and changed the names of string representations to something a little bit more concrete. 

@sparkprime Can we merge this?

PR #261, opened 2019-03-07
Fix improper stack trimming with operators and tailstrict

sbarzowski:
My understanding of origin of this bug was that once we
were creating thunks for binary operator arguments. So the environment
was no longer needed once they were created, so they could be tailcalls.

Now we're calling i.evaluate directly (for performance) and now
the environment cannot be destroyed during the evaluation of the first
argument.


sbarzowski:
@sparkprime 

I found this when working on #618.

sparkprime:
Those test cases would be good to add to test_suite/recursive_function.jsonnet

PR #262, opened 2019-03-08
Builtin implementation for std.sort

sbarzowski:
Sort is something that is highly optimized in most languages
and users can expect it to be fast. We can piggyback on
the Go implementation.

This change results in 100x speedup on `bench.06.jsonnet`.


sbarzowski:
Two important things are still missing:
a) Support for optional arguments in builtins - right now it expects keyF to always be present
b) Error handling when comparing values

sbarzowski:
Ready for review.

It turned out to be quite painful. I could have created something like `std.sortInternal(arr, keyF)` and avoid the need for optional arg, but I find this approach quite hacky (I think ideally everything possible in "real" Jsonnet should be implementable at a builtin level).

builtins.go
```diff
@@ -750,43 +834,6 @@ func builtinPow(i *interpreter, trace TraceElement, basev value, expv value) (va
 	return makeDoubleCheck(i, trace, math.Pow(base.value, exp.value))
 }
 
-func builtinSplitLimit(i *interpreter, trace TraceElement, strv, cv, maxSplitsV value) (value, error) {
```
sparkprime: Why drop this?

builtins.go
```diff
@@ -371,6 +376,85 @@ func builtinFilter(i *interpreter, trace TraceElement, funcv, arrv value) (value
...
+	if err != nil {
+		return nil, err
+	}
+	num := arr.length()
+
+	data := sortData{i: i, trace: trace, values: make([]value, num), keys: make([]value, num)}
+
+	for counter := 0; counter < num; counter++ {
+		var err error
+		data.values[counter], err = arr.elements[counter].getValue(i, trace)
```
sparkprime: We need to document a minor change in behavior in the release:
```jsonnet
std.length(std.sort([error 'lol', error 'lol2'], function(key)0))
```
Will now error out as the thunks are always evaluated.

sparkprime:
LGTM, it'd be good to get this in before the next release

builtins.go
sbarzowski: eh, looks like a rebase artifact, thanks for catching it

builtins.go
sbarzowski: Hmmm... perhaps there is no reason to evaluate them, since these are only swapped and never actually compared.

Generally though, I think that what parts of arguments are evaluated should be considered an implementation detail which may change freely from version to version, unless the documentation is very specific (which it isn't in this case). Even minor changes to the algorithm, data validation, little optimizations can change what exactly gets evaluated. And it's a bit tricky to test, too.

sparkprime: It'd be easier to test if there was in general a way to test for errors.  But that's not coming any time soon.

builtins.go
sbarzowski: fixed

builtins.go
sbarzowski: Fixed:
```
$ ./jsonnet -e "std.length(std.sort([error 'lol', error 'lol2'], function(key)0))"
2
```

PR #263, opened 2019-03-08
Keep object locals only once in AST

sbarzowski:
For example this reduces the size of stdlib ast file
roughly 3x. Note that this change doesn't regenerate the stdlib,
so that the diff here is sane.

It is likely to slightly improve performance of code using
a lot of locals (~10% on bench.05.gen.jsonnet).

The desugaring is more strightforward now, and we're back
to desugaring each node exactly once.


sbarzowski:
I found out that it leads to massive performance improvement (~4x) on loading k8s.libsonnet. 90% of evaluation time was spent desugaring and analyzing the blown up AST. 

@sparkprime

desugarer.go
```diff
@@ -182,13 +196,18 @@ func buildAnd(left ast.Node, right ast.Node) ast.Node {
...
-func desugarForSpec(inside ast.Node, forSpec *ast.ForSpec) (ast.Node, error) {
+// inside is assumed to be already desugared (and cannot be desugared again)
+func desugarForSpec(inside ast.Node, forSpec *ast.ForSpec, objLevel int) (ast.Node, error) {
 	var body ast.Node
 	if len(forSpec.Conditions) > 0 {
 		cond := forSpec.Conditions[0].Expr
 		for i := 1; i < len(forSpec.Conditions); i++ {
 			cond = buildAnd(cond, forSpec.Conditions[i].Expr)
 		}
+		err := desugar(&cond, objLevel)
```
sparkprime: Did cond and forSpec.Expr get forgotten about before?

desugarer.go
```diff
@@ -253,7 +270,7 @@ func buildLiteralString(value string) ast.Node {
 func buildSimpleIndex(obj ast.Node, member ast.Identifier) ast.Node {
 	return &ast.Index{
 		Target: obj,
-		Id:     &member,
+		Index:  buildLiteralString(string(member)),
```
sparkprime: ast.Identifier is immutable though isn't it?

value.go
```diff
@@ -456,6 +456,13 @@ func (obj *valueObjectBase) getAssertionsCheckResult() error {
 	return obj.assertionError
 }
 
+type objectLocal struct {
+	name ast.Identifier
+	// Locals may depend on self and super so they are unbound fields and not simply thunks
+	// uf unboundField
```
sparkprime: line is commented out

desugarer.go
sbarzowski: No, it was just desugared in two steps. First the forSpec was desugared then the result was desugared again, now without the comprehensions, which was clearly not ideal (both for performance and our sanity).

desugarer.go
sbarzowski: The underlying data of strings (and `ast.Identifier` which is the same thing underneath) is immutable. But it's possible to do something like `a := "foo"; a = "bar"`. 
 
The `string(member)` is there because `string` and `ast.Identifier` need to be explicitly converted, even though they are really the same thing underneath.

(And the point of changing this line is to build desugared node in one step, rather than something that needs to be desugared).

sparkprime:
LGTM

Issue #126, opened 2017-10-10
Consider exposing interface for repeated execution with varying environment


seh:
At present, it appears that _go-jsonnet_'s public interface starts with `jsonnet.MakeVM`, against which one can call `(*VM).ExtVar`, `(*VM).ExtCode`, and `(*VM).Importer` to prepare an environment, and eventually call `(*VM).EvaluateSnippet` against a sequence of Jsonnet expressions.

That call to `(*VM).EvaluateSnippet` eventually calls on the internal `snippetToAST` function, which yields a parsed `ast.Node` ready for later use in the `evaluate` function. In there, we find some more code-to-AST translation for environment entries that are Jsonnet code.

Is it viable to cleave a new break into this interface, so that it's possible to parse some Jsonnet—together with its import callbacks, and perhaps code entries from the environment—and yield a prepared object that could then be used for repeated evaluation against different TLAs?

The use case I have in mind is a server that would take a Jsonnet function that expects to receive some JSON as a TLA and evaluate the same Jsonnet function repeatedly and frequently, just with its input changing. It's unfortunate to have to restart this procedure each time from raw text when the code involved won't change between evaluations, throwing away the parsed form of the input text.

I concede that I have no measurements to critique the cost of the current interface compared to the separate prepare-once/evaluate-repeatedly approach sketched above. Has this project considered and rejected this approach before? If not, I'm willing to help pursue it.

A few ideas for an interface:

- `(*vm).PrepareSnippet(filename string, snippet string) (*Program, error)`
- `(*Program).Eval(vs map[string]string) (string, error)`
- `(*Program).EvalTo(w io.Writer, vs map[string]string) error`

sbarzowski:
I like the interface you proposed. I was thinking about something similar that it in the context of #82 (but there the idea was to reuse the whole interpreter or parsed std object with different programs).

seh:
I'm sure I'm making some rookie mistake here, failing to take the late-bound processing requirements into account, and I don't want to propose something that limits how the language works.

I came to this idea after fiddling around with the _jsonnet_ command-line tool for a while, where I was feeding in various JSON files via the `--tla-code-file` flag to the same Jsonnet expression (starting with `function(...)`) repeatedly. I'm not sure whether using the `--tla-str-file` flag would have worked just as well, given that the input content was JSON, and not Jsonnet per se.

sbarzowski:
[...]

PR #264, opened 2019-03-08
Expose Evaluate functions, to be able to reuse the AST from `SnippetToAST`

abourget:
Right now, SnippetToAST provides a `Node` but nothing consumes it it seems.

This is mostly to open the discussion. My goal is to use jsonnet as a reformatter in streaming objects.. Performance is going to be paramount for us, reusing the AST is a start.

Nice work in here by the way.. love the project :)

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -96,6 +96,33 @@ const (
 	evalKindStream           = iota
 )
 
+func (vm *VM) Evaluate(node ast.Node) (output interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


seh:
This looks like it might be related to #126.

abourget:
It is indeed :)

sparkprime:
@sbarzowski seems ok to me, how about you?

sbarzowski:
LGTM, since we are exposing AST api and evaluation api, I don't see any reason not to expose AST evaluation API :-)

One day, we may want to extend and clean up that API, with efficient support for evaluating multiple file at once, or the same one multiple times with different TLAs - preserving cached values, especially for imported libs.

PR #266, opened 2019-03-19
Add go mod / cleanup

michaeljs1990:
I can break this up into two diffs if you want but I made the go mod change in order to make the build easier which then lead me to make the other changes.

This adds the new go dependency management tooling seen here https://github.com/golang/go/wiki/Modules. You can now easily clone this and build it any place you want. Additionally I changed the format for building the binaries to be moved under ./cmd/${dirname} so when using go build you no longer have to cd into the directory to build it or set the -o flag.


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/google/go-jsonnet
+
+require (
+	github.com/fatih/color v1.7.0
```
sbarzowski: At some point we got complaints that we depend on `color` even for people who use go-jsonnet as a library only. Is there a way to have separate dependencies for the library and the command?

go.sum
```diff
@@ -0,0 +1,18 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
sbarzowski: How did we end up with go-spew here? It's very useful for debugging, but I don't think we have an explicit dependency in the code...

README.md
```diff
@@ -16,21 +16,18 @@ feedback.
 
 This code is known to work on Go 1.8 and above. We recommend always using the newest stable release of Go.
```
sbarzowski: Do we need to bump supported Go version?

sbarzowski:
Thanks! This is definitely what we should do.

README.md
michaeljs1990: This will still work with go 1.8 but you are correct that the git clone and build workflow will not work and if we want to maintain both I can add docs for 1.8 vs 1.11+.

go.sum
michaeljs1990: The go.sum file came from running `go mod tidy` we can manually clean this up but it will keep getting added back when issuing commands like `go build` or `go get` since it will see the package and include it.

https://github.com/golang/go/issues/26913#issuecomment-411976222 covers the current state of build vs test dependencies and the hopeful fixes that it should get in the future.

go.mod
sbarzowski: ok, it's good then

go.sum
sbarzowski: Oh, it comes from tests. That makes sense. It doesn't really matter - if I understand correctly `go.sum` is just used for verification that the dependencies are legitimate.

README.md
sbarzowski: Yeah, if you could do that, it would be nice. Go 1.10 is the default version in Ubuntu for example, so I think it would be better to also have instructions covering Go <1.11.

michaeljs1990: All set, I added instructions for 1.8-1.10 and 1.11+.

sbarzowski:
LGTM

@sparkprime

PR #268, opened 2019-04-11
Add merge sort for sorting large arrays

michaeljs1990:
See https://github.com/google/jsonnet/pull/619 for more info. As this is just copy and pasted from your other pull request please feel free to steal and put your name on it. I just wanted to be able to run my code under the golang version of jsonnet which requires merge sort.

If you would like the tests ported over as well I can look into it but it seems like the specific C++ ones that were touched aren't in the golang repo right now.

### DIFF
--- a/std/std.jsonnet
+++ b/std/std.jsonnet
@@ -1132,17 +1132,41 @@ limitations under the License.
     local bytes = std.base64DecodeBytes(str);
     std.join('', std.map(function(b) std.char(b), bytes)),
 
-  // Quicksort
+  // Merge-sort for long arrays and naive quicksort for shorter ones
   sort(arr, keyF=id)::
+    local quickSort(arr, keyF=id) = 
+      local l = std.length(arr);
+      if std.length(arr) <= 1 then
+        arr
+      else
+        local pos = 0;
+        local pivot = keyF(arr[pos]);
+        local rest = std.makeArray(l - 1, function(i) if i < pos then arr[i] else arr[i + 1]);
+        local left = std.filter(function(x) keyF(x) < pivot, rest);
+        local right = std.filter(function(x) keyF(x) >= pivot, rest);
+        quickSort(left, keyF) + [arr[pos]] + quickSort(right, keyF);
+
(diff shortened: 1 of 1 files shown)


sbarzowski:
Thanks. I created a new change (#269) which syncs std.jsonnet with cpp-jsonnet (it's supposed to be the same) and makes it a little bit easier to keep them in sync in the future.

FYI when you're modifying stdlib for golang version you need to regenerate some code like this:
```
./reset_stdast_go.sh && go run cmd/dumpstdlibast/dumpstdlibast.go
```
Otherwise it has no effect.

Anyway it's all done in #269.

phatlast96:
Why is the standard library written in itself? 

https://databricks.com/blog/2018/10/12/writing-a-faster-jsonnet-compiler.html

michaeljs1990:
@sbarzowski thanks for updating that!

sbarzowski:
@phatlast96 You mean why the standard library is written in Jsonnet rather than implemented natively in the language of the interpreter? 

First of all there approach is actually hybrid - some function is stdlib have native implementations. Some functions have both (and the native one is overriding the one in Jsonnet).

A bunch of practical reasons:
* It allows moving simply promoting Jsonnet code to standard library
* It can be used in any implementation of the interpreter (especially important for complicated/less obvious stuff like string formatting)
* It provides a reference implementation
* Something being reasonably implementable in Jsonnet is a good sign for being a good fit for stdlib (e.g. no side effects)
* It's easier to write these functions in Jsonnet than as builtins (especially in cpp-jsonnet, in go-jsonnet it's mostly just more boilerplate).
* It was done like that in the beginning

BTW there are definitely some functions from stdlib which deserve native implementations (e.g. substr, map, foldl, foldr, abs, sign, max, min). Adding one or more of them is a good candidate for the first contribution (see `builtins.go`) if you are interested in helping with that.

phatlast96:
@sbarzowski  Thank you so much for taking the time out of your busy day to write this. Much appreciated!! I'll see what I can do.

Issue #205, opened 2018-03-02
Reuse C++ Jsonnet's std.jsonnet


marcelocantos:
#202 Brings the C++ Jsonnet into go-jsonnet as a submodule. Consequently, its version of std.jsonnet is readily available at a known path.

Should `cpp-jsonnet/stdlib/std.jsonnet` be made the common source of truth for the standard library?

marcelocantos:
The main caveat with this approach is that submodules don't update by default when using git's cli (though I believe most GUI-based git clients do). This carries a small risk that a developer forgets to `git submodule update` and spends a good half hour confused and frustrated. On the plus side, it wouldn't be an issue with the standard instructions in the README, since go get automatically fetches submodules (https://github.com/golang/go/commit/a444da033c0bc1fa5fb1973bd6d0c131ec3c1ccd).

sparkprime:
If this can work smoothly with existing tooling, we should definitely do it.

sbarzowski:
Fixed by #269

PR #269, opened 2019-04-12
Share std.jsonnet with cpp-jsonnet

sbarzowski:
So far `std.jsonnet` needed to be updated seprately from
the cpp-jsonnet submodule. Since we should update it anyway
at the same time (to make sure the tests are running fine),
we can just as well get it directly from there, eliminating
the extra step of copying the new `std.jsonnet` version.

This change updates the cpp-jsonnet used.


sbarzowski:
@sparkprime @michaeljs1990

sbarzowski:
This is basically a sync which fixes an issue + minor restructuring we wanted to do for a long time (https://github.com/google/go-jsonnet/issues/205), so I'm just going to move forward and merge this.

PR #280, opened 2019-05-25
Object field caching

sbarzowski:
So far I did this in a way which is intended to be minimally
invasive, even at the cost of worse performance and some weirdness.
Most importantly valueCachedObject should probably replace valueObject
interface to avoid some indirection (and it shouldn't be at the same
"level" as valueSimpleObject/valueExtendedObject (these should no longer
even be values on their own).

This is a reasonable proof of concept, which we can try benchmarking
on the real world code.

### DIFF
--- a/value.go
+++ b/value.go
@@ -357,16 +357,21 @@ func args(xs ...*cachedThunk) callArguments {
 
 // Object is a value that allows indexing (taking a value of a field)
 // and combining through mixin inheritence (operator +).
-//
-// Note that every time a field is indexed it evaluates it again, there is
-// no caching of field values. See: https://github.com/google/go-jsonnet/issues/113
-type valueObject interface {
-	value
-	inheritanceSize() int
-	index(i *interpreter, trace TraceElement, field string) (value, error)
-	assertionsChecked() bool
-	setAssertionsCheckResult(err error)
-	getAssertionsCheckResult() error
+
+type valueObject struct {
+	valueBase
+	assertionError error
+	cache          map[objectCacheKey]value
+	uncached     uncachedObject
+}
+
+// Hack - we need to distinguish not-checked-yet and no error situations
+// so we have a special value for no error and nil means that we don't know yet.
+var errNoErrorInObjectInvariants = errors.New("no error - assertions passed")
+
+type objectCacheKey struct {
+	field string
+	depth int
 }
 
 type selfBinding struct {
@@ -374,7 +379,7 @@ type selfBinding struct {
 	// that this is not the same as context, because we could be inside a function,
 	// inside an object and then context would be the function, but self would still point
 	// to the object.
-	self valueObject
+	self *valueObject
 
 	// superDepth is the "super" level of self.  Sometimes, we look upwards in the
 	// inheritance tree, e.g. via an explicit use of super, or because a given field
@@ -395,7 +400,7 @@ func makeUnboundSelfBinding() selfBinding {
 	}
 }
 
-func objectBinding(obj valueObject) selfBinding {
+func objectBinding(obj *valueObject) selfBinding {
 	return selfBinding{self: obj, superDepth: 0}
 }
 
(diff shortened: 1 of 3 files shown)


sbarzowski:
Related: https://github.com/google/jsonnet/pull/663

I think we can do object local caching in another change after this gets finished.

davidreuss:
We have a fairly large corpus of configuration that takes ~5s to render -- I read the article on `sjsonnet` (https://databricks.com/blog/2018/10/12/writing-a-faster-jsonnet-compiler.html), and realized that object caching might be what is causing our configuration to be slow to render.

We got a few native functions so we can't apply `sjsonnet` directly as a measure, but .. there's very real benefits with this patch for us so, that's interesting to pursue.

We go from ~4.5s to about ~2.8s with this patch applied, and it makes a noticeably difference when "developing" on the config, and trying to expand the entire thing.

Is there anything hindering this from being merged?

sparkprime:
Thanks for describing your situation. I haven't reviewed it yet but it's high up my priorities.  I don't suppose your config is public is it? :)

davidreuss:
No it’s not — it’s non prod sensitive though (it’s generating a complete environment for our ci/e2e testing) so we can possibly figure something out if it would be of value in determining if these perf changes are working on a non-trivial config

we do have a couple of custom functions and currently we’re using an in-house wrapper for that but we can maybe figure something out if you’re interested .. might also be able to strip/customize some specific bits so a native jsonnet evaluator could be used ..

sparkprime:
I suspect it'd be possible to swap out the native bits with hardcoded values or jsonnet functions without changing the performance of it

sparkprime:
Also might be possible to anonymise it by changing all the strings to "DDDDDDDDDDDDD" or whatever

davidreuss:
i’ll get back to you when i have something simplified to try things out with :)

sbarzowski:
FYI I expect to finish the refactoring and drop WIP mark today or tomorrow.

davidreuss:
I've put together a container which allows expanding our configuration with plain command line 

I tried just for kicks both with the current `go` version and with the `c++` version ... 

*c++*

```
$ time ./expand.sh >/dev/null
./expand.sh > /dev/null  24.18s user 1.19s system 99% cpu 25.562 total
```

*go*

```
$ time ./expand.sh >/dev/null
./expand.sh > /dev/null  4.07s user 0.20s system 139% cpu 3.069 total
```

value.go
```diff
@@ -357,24 +357,29 @@ func args(xs ...*cachedThunk) callArguments {
...
-	index(i *interpreter, trace TraceElement, field string) (value, error)
-	assertionsChecked() bool
-	setAssertionsCheckResult(err error)
-	getAssertionsCheckResult() error
+
+type valueObject struct {
+	valueBase
+	assertionError error
+	cache          map[objectCacheKey]value
+	underlying     objectBody
```
sbarzowski: Any ideas for a better names than "underlying" and "objectBody"?

value.go
```diff
@@ -647,10 +657,20 @@ func objectIndex(i *interpreter, trace TraceElement, sb selfBinding, fieldName s
 		return nil, i.Error(fmt.Sprintf("Field does not exist: %s", fieldName), trace)
 	}
 
+	if val, ok := sb.self.cache[objectCacheKey{field: fieldName, depth: foundAt}]; ok {
```
sbarzowski: Here, in this function the actual caching happens.

sbarzowski:
Ready for review

davidreuss:
I’ll try and apply this tomorrow and test it out... i’ll also see if i can anonymize our code so you can see for yourselves, if you got time for it

sbarzowski:
@sparkprime Are we going forward with this?

sparkprime:
Yeah I started reviewing it last week

value.go
sparkprime: We could rename the top-level one valueCachedObject and then the other one could just be object or valueObject or uncachedObject.  I don't really like the word "body" as it usually implies the bit that isn't a header but there's no header here.

value.go
```diff
@@ -419,34 +424,33 @@ func withHiddenFromBool(with bool) Hidden {
...
-type valueObjectBase struct {
-	valueBase
-	assertionError error
+func (o *valueObject) index(i *interpreter, trace TraceElement, field string) (value, error) {
+	return objectIndex(i, trace, objectBinding(o), field)
 }
 
-func (*valueObjectBase) getType() *valueType {
-	return objectType
+func (obj *valueObject) inheritanceSize() int {
```
sparkprime: Does this ever get called on the top level (or if it does, do you happen to have the inside object readily available)?

sparkprime:
LGTM

value.go
sbarzowski: It does, but only in context where we are playing with the object internals anyway, so we can reach for the uncachedObject. I removed it.

It comes with an added advantage - valueObject no longer conforms to uncachedObject interface - so they cannot be passed there by mistake (and there were some places where it was unnecessarily passed instead of uncachedObject). It is now fixed. 

Thanks for noticing this.

value.go
sbarzowski: The `objectBody`, so valueObject is out of the question. I like the `uncachedObject` name the most.

So I renamed `objectBody` to `uncachedObject` and `underlying` to `uncached`. I left `valueObject` as it was, because I think it's clear enough - it is the only representation of object values (uncachedObject is not a value on its own).

sparkprime:
Alright, merge it when you're ready. Nice job!

PR #283, opened 2019-05-29
Fix unintentionally invalid filename escaping test

sbarzowski:


seh:
See #285 for a possibly conflicting change.

PR #285, opened 2019-06-03
golden files were out of date

sparkprime:
@sbarzowski Any idea why CI didn't pick this up?

### DIFF
--- a/testdata/import_various_literals_escaped.golden
+++ b/testdata/import_various_literals_escaped.golden
@@ -1,6 +1,6 @@
 [
    true,
    true,
-   "// This file is there only for its filename: to test escaping in imports\n{}\n",
-   "// This file is there only for its filename: to test escaping in imports\n{}\n"
+   "// This file is there only for its filename: to test escaping in imports\n",
+   "// This file is there only for its filename: to test escaping in imports\n"
 ]
--- a/testdata/importstr_various_literals_escaped.golden
+++ b/testdata/importstr_various_literals_escaped.golden
@@ -1,6 +1,6 @@
 [
    "true\n",
    "true\n",
-   "// This file is there only for its filename: to test escaping in imports\n{}\n",
-   "// This file is there only for its filename: to test escaping in imports\n{}\n"
+   "// This file is there only for its filename: to test escaping in imports\n",
+   "// This file is there only for its filename: to test escaping in imports\n"
 ]


seh:
There was recent discussion [here](https://github.com/google/go-jsonnet/pull/281#issuecomment-496686229) and [here](https://github.com/google/go-jsonnet/pull/281#issuecomment-497080696), thought to be fixed in #283.

sbarzowski:
@sparkprime Are you sure they were out of date? The tests seem to be failing now after this change was merged.

But yeah, CI didn't fail when it encountered the problem (even though from what I see the problem is in this change, not in #283, see https://travis-ci.org/google/go-jsonnet/jobs/540843033). I think it's because we don't catch the error on this line  https://github.com/google/go-jsonnet/blob/v0.13.0/travisBuild.sh#L4.

sparkprime:
`go test` wasn't working on my Linux laptop. I looked at the jsonnet and golden files and it looked like `{}` got removed from the jsonnet files (and they were imported with `importstr` instead of `import`) but the golden file remained the same.

However, given the other weirdness that I just read about in linked discussions, I'm not sure anymore.

sparkprime:
I can't make heads or tails of that test failure, i guess it was transient?

sparkprime:
The CI failure I mean.

sbarzowski:
> it looked like {} got removed from the jsonnet files

But #283 added `{}` to these files. Perhaps somehow they got overwritten on your machine? This is weird. Anyway right now the tests don't pass on master for me (on Linux and on Mac). 

Good thing #285 gets rid of escaping business.

sbarzowski:
And the CI failure seems like it has something to do with running from non-master branch. We had the same thing with v0.12.1 and v0.12.0 - the same code was passing the tests on master, but failed on other branches.

sparkprime:
They pass for me on master and Linux right now. I wonder if we have an ordering problem or maybe something that is .gitignored but is affecting behavior?  I could try a fresh clone but I'm on a plane still and it might barf at the amount of data

sparkprime:
@sbarzowski is this the failure you're reporting (the overall test is still passing, which is wrong presumably) https://travis-ci.org/google/go-jsonnet/jobs/540870045  I'm very confused :)

sparkprime:
I see my `".jsonnet` and `'.jsonnet` files were out of date, they were ignored by testdata/.gitignore and it wasn't regenerating them for some reason, I guess because they already existed and there is no code to look at the date of the files.

PR #287, opened 2019-06-04
Don't assume Go is in a specific directory

sparkprime:
@sbarzowski I tried to run travisBuild.sh locally and ran into this.  Not sure this is the correct fix though.


sbarzowski:
Not sure if it's The Right Way™, but it's definitely an improvement, so LGTM.

PR #288, opened 2019-06-04
Change golden files back again;

sparkprime:
Roll back #285


seh:
These now match the content pending in #281. See the `TestEvalUnusualFilenames` function in file _main_test.go_.

PR #290, opened 2019-06-21
Eliminate Bazel local repository for "cpp-jsonnet" submodule

seh:
Treating the "cpp-jsonnet" Git submodule as a separate local Bazel repository precludes use of the "go-jsonnet" repository from other workspaces, as Bazel misinterprets the relative path to the local
"cpp-jsonnet" repository. Instead, use the Git submodule content directly along the package path from the workspace root.

Thanks to @werkt for pointing out the root of the problem.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,10 +1,5 @@
 workspace(name = "google_jsonnet_go")
 
-local_repository(
-    name = "cpp_jsonnet",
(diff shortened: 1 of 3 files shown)


sparkprime:
Seems reasonable, thanks

seh:
See bazelbuild/rules_jsonnet#113 for a proposed use of these Bazel targets.

PR #291, opened 2019-06-22
Fix double desugaring in array comps

sbarzowski:
Accidentally the array we were iterating over was desugared
two times, which resulted in crash when an object was there.


sbarzowski:
@seh it should fix your problem reported in #290.

@sm-gravid-day It's very likely that that it's the same issue as the crash reported here https://github.com/google/jsonnet/issues/674.

PR #292, opened 2019-07-02
Allow use of "go-jsonnet" Bazel repository as an external dependency

seh:
In order to rely on the _go-jsonnet_ Bazel repository as an external dependency from other Bazel projects, until [it becomes possible to load a _WORKSPACE_ file](https://github.com/bazelbuild/bazel/issues/1943), it's necessary to first declare the repositories on which this repository depends (_rules_go_ and _gazelle_), and then to load file from those two repositories and call macros to initialize them.

Expose two files to handle these steps:
- _bazel/repositories.bzl_ . 
  Defines the `jsonnet_go_repositories` macro.
- _bazel/deps.bzl_ . 
  Defines the `jsonnet_go_dependencies` macro.

Use these files and the macros they define within this repository's _WORKSPACE_ file, as a dependent repository would do in its _WORKSPACE_ file.


sbarzowski:
I don't get the full Bazelology, but it looks reasonable.

PR #296, opened 2019-07-21
Make golint happy

sbarzowski:


seh:
I'm happy to see fewer exported identifiers.

I don't know if this horse has already sailed, but have you considered introducing an _internal_ directory tree for packages that you don't want anyone relying on from outside the project? This project serves at present as both the source for a tool and a library, but it makes little distinction above the individual package level which parts are meant for external use.

sbarzowski:
Actually the reason we had unnecessarily exported identifiers was that when I was writing the first versions of this interpreter I expected to split it into many packages and I wanted to avoid renaming. In retrospect it was pretty stupid on my part and the "temporary" random exports persisted for a long time. Anyway there are just a few random things left and I intend to get rid of them soon.

In foreseeable future I think the only candidates for splitting into a separate (probably internal) packages are desugaring and static analysis passes.

seh:
As an example, while working on #281, I had to export a bunch of struct types and their fields in the `ast` package. No one objected, but it wasn't clear to me whether this package is an internal detail, or whether it's supposed to be available to "library users." I know the answer now (it's part of the public interface), but at the time, I thought that if the package had been within an _internal_ directory, I could have made that change without worrying whether I was exposing too much, creating a liability by restricting how we could change it in future.

The more we export, the less we can change without risk of breaking existing programs using Jsonnet as a library.

sparkprime:
I agree it should be clear what interfaces are intended for public consumption.

PR #297, opened 2019-07-21
Enforce no golint warnings in CI

sbarzowski:


sbarzowski:
For now I'm just experimenting with it.

sbarzowski:
Ok, it seems to work. I also think it solves the unfortunate old issue with "tests failing but being marked as passed" (due to `set -e`). 

The downside is that we no longer test on Go 1.8.x - it's too old for golint. I think we should just upgrade 1.10 is old enough (there is a separate commit in this PR for that).

sbarzowski:
Of course first we need to finish #296

sbarzowski:
I noticed that it only checks Jsonnet package this way (and not `ast` etc.). This needs to be fixed.

sbarzowski:
Hmmm... also perhaps https://github.com/golangci/golangci-lint is a better tool for the job, mostly because golint doesn't have features for suppressing warnings (and sometimes it is just wrong or interfering with backwards compatibility, so if we force lint-free code, we need a way to locally sidestep it).

sbarzowski:
Superseded by #313.

Issue #299, opened 2019-07-23
Failure to render kube-prometheus jsonnet on CircleCI


thejosephstevens:
Running jsonnet 0.13.0

After running this on CircleCI
```
FATAL: Command 'jsonnet -J monitoring/kube-prometheus/vendor -m monitoring/kube-prometheus/manifests-generated/json --tla-str 'Host=myHostName' --tla-str 'environment=gcp' monitoring/kube-prometheus/compute.jsonnet' returned non-zero exit status 1.
```

This error was output. I'm unable to reproduce the issue locally and I have not been able to identify a version I can switch that will fix it (have played with kube-prometheus version, have tried jsonnet 0.12.1, have tried bumping etcd-mixin to master)
```
INTERNAL ERROR: (CRASH) runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack(0xc008d185d8, 0x5609c0, 0x6e3c00)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9d
github.com/google/go-jsonnet.(*VM).evaluateSnippet.func1(0xc008d21d00)
	/go/src/github.com/google/go-jsonnet/vm.go:129 +0x60
panic(0x5609c0, 0x6e3c00)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/google/go-jsonnet.makeValueExtendedObject(0xc008770090, 0xc008755920, 0x1)
	/go/src/github.com/google/go-jsonnet/value.go:585 +0x5d
github.com/google/go-jsonnet.builtinPlus(0xc0000cc140, 0xc0001fdf00, 0xc0001690c0, 0x5dcde0, 0xc008770090, 0x5dcde0, 0xc008755920, 0x5dcde0, 0xc008770090, 0x0, ...)
	/go/src/github.com/google/go-jsonnet/builtins.go:61 +0x174
github.com/google/go-jsonnet.(*PlusSuperUnboundField).evaluate(0xc0005b8780, 0xc0000cc140, 0xc0001fdf00, 0xc0001690c0, 0xc0007053e0, 0x0, 0xc0087555f0, 0xc0045083a0, 0x11, 0xc0007052f0, ...)
	/go/src/github.com/google/go-jsonnet/thunks.go:139 +0x216
[...]

PR #301, opened 2019-07-24
Fix crash when using empty object comprehension

sbarzowski:
The object was not properly created when object comprehension
was empty.


sbarzowski:
Fixes #299

Eh, I screwed up the object comprehensions again in #280 :-(. This time at least the problem is shallow.

@sparkprime

PR #304, opened 2019-07-26
Extract AST processing to separate packages

sbarzowski:
Making it independent from the jsonnet package breaks the circular
dependency during stdast generation.


internal/transformations/desugarer_test.go
```diff
@@ -14,4 +14,4 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package jsonnet
+package transformations
```
sbarzowski: I'm open to suggestions for a better package name. The rationale behind "transformations" is that it transforms the AST.

cmd/dumpstdlibast/dumpstdlibast.go
```diff
@@ -19,8 +19,8 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/google/go-jsonnet"
-	"github.com/google/go-jsonnet/dump"
+	"github.com/google/go-jsonnet/internal/dump"
+	"github.com/google/go-jsonnet/internal/transformations"
```
sbarzowski: Here we avoid the dependency on jsonnet package.

parser/parser.go
```diff
@@ -1191,3 +1191,12 @@ func Parse(t Tokens) (ast.Node, error) {
 
 	return expr, nil
 }
+
```
sbarzowski: I guess `parser` should also be internal, but I want to move it later in a separate change that cleans up all remaining unnecessarily exported stuff.

sbarzowski:
@seh @sparkprime

internal/transformations/transformations.go
```diff
@@ -0,0 +1,31 @@
...
+
+func snippetToRawAST(filename string, snippet string) (ast.Node, error) {
+	tokens, err := parser.Lex(filename, snippet)
+	if err != nil {
+		return nil, err
+	}
+	return parser.Parse(tokens)
+}
+
+// SnippetToAST converts Jsonnet code snippet to desugared and analyzed AST
```
seh: This sentence is missing a word or two. Perhaps:
> SnippetToAST converts a Jsonnet code snippet to a desugared and analyzed AST.

parser/parser.go
```diff
@@ -1191,3 +1191,12 @@ func Parse(t Tokens) (ast.Node, error) {
...
+// SnippetToRawAST converts Jsonnet code snippet to AST (without any transformations)
```
seh: This sentence is missing a word or two. Perhaps:
> SnippetToRawAST converts a Jsonnet code snippet to an AST (without any transformations).

README.md
```diff
@@ -47,6 +47,11 @@ Windows         | _bazel build --platforms=@io_bazel_rules_go//go/toolchain:wind
 
 For additional target platform names, see the per-Go release definitions [here](https://github.com/bazelbuild/rules_go/blob/master/go/private/sdk_list.bzl#L21-L31) in the _rules_go_ Bazel package.
 
+Additionally if files were moved around, you may need to run the following command to update Bazel files:
```
seh: We do mention this point below in the "Keeping the Bazel files up to date" section. I just want to confirm that you intended to duplicate this point here.

c-bindings/BUILD.bazel
```diff
@@ -7,6 +7,7 @@ go_library(
         "internal.h",
         "json.cpp",
         "json.h",
+        "libgojsonnet.h",
```
seh: Please see [my comment from #281](https://github.com/google/go-jsonnet/pull/281#issuecomment-497971532) about this entry. I think we should leave it out here, though I still don't know of a way to keep it from coming back.

internal/transformations/desugarer_test.go
seh: Perhaps just "transform" would read better for importing code: `transform.SnippetToRawAST`.

internal/transformations/static_analyzer.go
```diff
@@ -164,6 +164,9 @@ func analyzeVisit(a ast.Node, inObject bool, vars ast.IdentifierSet) error {
 	return s.err
 }
 
-func analyze(node ast.Node) error {
+// Analyze checks variable references (these could be checked statically in Jsonnet).
+// It enriches ast with additional information about free variables in every node,
+// so it is necessary to always run it before executing AST.
+func Analyze(node ast.Node) error {
```
seh: That this function is exported from this package suggests that neither "transformations" nor "transform" capture all of what's going on in here. The documentation suggests that this function mutates the supplied AST in place.

Given that one is not supposed to execute an AST until it's been analyzed, we could enforce that by having this function return a type that tolerates such execution, and disallow doing that directly against a raw AST. Unfortunately, that would break backward compatibility.

Instead of offering `(*VM).Evaluate(ast.Node)` and kin, we could call this package "program" introduce a type like `program.Top` that wraps a prepared `ast.Node`.
```go
func Prepare(node ast.Node) (Top, error) {
```
We'd then rephrase `(*VM).Evaluate(ast.Node)` and kin as `(*VM).Evaluate(program.Top)`. In other words, you couldn't feed an `ast.Node` into the evaluation procedures, because it hadn't been properly prepared.

seh: Does anything call on this function from outside the package? Do we need to export it?

cmd/dumpstdlibast/dumpstdlibast.go
```diff
@@ -33,7 +33,7 @@ func main() {
 		panic(err)
 	}
 
-	node, err := jsonnet.SnippetToAST("<std>", string(buf))
+	node, err := transformations.SnippetToAST("<std>", string(buf))
 	if err != nil {
 		panic(err)
 	}
```
seh: Since I can't leave a comment on a line that didn't change, I'll ask as close by as possible: Is it correct that you left the `init` function in the `header` value here for now with the intention to come through in a second pass and remove the duplicated installation of the standard library AST?

seh:
With these changes, it looks like we could eliminate the "astgen" package, and generate the standard library AST directly into the "jsonnet" package—if that would help. The "astgen" package depends on the "ast" package, but the "ast" package does not depend on the "jsonnet" package.

What do you think about doing that?

README.md
sbarzowski: I added this because I was a bit confused before I figured out the gazelle thing. I didn't really think about the other section. I changed the text so that it references the other section.

c-bindings/BUILD.bazel
sbarzowski: Hmmm..., this way, it's only a matter of time until it slips back. I created a separate PR #308 which hopefully will keep it out.

internal/transformations/static_analyzer.go
sbarzowski: Hmmm... I like the name "program". Even without introducing a new type it makes sense (it prepares the program to be run). It's also short and nice.

For now I'm leaving out the new type but I agree with the argument for it. It just seems out of scope here and it I'm not sure if backwards compatibility issues in this case will be acceptable.

sbarzowski: No, I unexported it. 

Originally Analyze and Desugar were supposed to be high-level operations here, and SnippetToAST was just a helper function in my original idea for this package. But since that's the only part that is actually used, it makes sense to unexport these two.

cmd/dumpstdlibast/dumpstdlibast.go
sbarzowski: I intend to do the whole astgen cleanup in separate PR. I like to keep the changes that move stuff around as small and focused as possible.

internal/transformations/desugarer.go
```diff
@@ -579,7 +571,18 @@ func desugar(astPtr *ast.Node, objLevel int) (err error) {
 	return nil
 }
 
-func desugarFile(ast *ast.Node) error {
+// Desugar Jsonnet expressions to reduce the number of constructs the rest of the implementation
+// needs to understand.
+//
+// Note that despite the name, desugar() is not idempotent.  String literals have their escape
```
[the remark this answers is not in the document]
sbarzowski: It will unescape it again. So `"\"blah\""` becomes `"blah"` after first unescaping and if we tried to unescape it again we would get `blah`. This is not very clean and perhaps we should just add a separate field for that. 

Anyway for performance reasons it's better not to run desugaring on the same nodes again and again.

internal/transformations/desugarer_test.go
sbarzowski: I went with `program`.

sbarzowski:
I'm going to merge this soon if no one objects.

PR #308, opened 2019-08-14
Prevent autogenerated header in BUILD.bazel

sbarzowski:
Before this change updating BUILD.bazel file with Gazelle
would add `c-bindings/libgojsonnet.h` to `srcs/`.

It was prevented manually, but someone is going to let it slip
in at some point, so it's better to just not include it there
in the first place.


sbarzowski:
In response to: 
* https://github.com/google/go-jsonnet/pull/281#issuecomment-497971532
* https://github.com/google/go-jsonnet/pull/304#discussion_r308236126

I don't really know much about Bazel and even less about Gazelle, but I run `bazel run //:gazelle` a few times and it seems `libgojsonnet.h` is no longer added to `srcs`.

@seh Do you think it's a reasonable solution?

seh:
> In response to:
> @seh Do you think it's a reasonable solution?

Yes, that looks like a fine way to cover this problem. I don't know if Gazelle would honor [the directive](https://github.com/bazelbuild/bazel-gazelle#directives) if we put it down in the _c-bindings/BUILD.bazel_ file, closer to the target, so that it would be more obvious what it's referring to. It may be the case that these exclusions only work when stated with paths relative to the root.

sbarzowski:
Oh, it worked in *c-bindings/BUILD.bazel*. I assumed directives cannot be placed there for some reason and I didn't even try that before. 

Thanks!

c-bindings/BUILD.bazel
```diff
@@ -1,5 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
+# gazelle:exclude ./libgojsonnet.h
```
seh: Consider writing a comment mentioning why this file will show up: namely, running _go build_ instead of _bazel build_. In the former case, cgo creates the file in the current directory. In the latter case, the file gets created elsewhere in Bazel's sandbox tree, so we never see it (unless you deliberately go looking around for it).

PR #310, opened 2019-09-01
Eliminate the unnecessary ast.StdAst

sbarzowski:
The initially empty ast.StdAst was created to break the circular
dependency. The generation of stdlib AST used to depend on
the primary "jsonnet" package, which meant that "jsonnet"
could not depend on it directly. Hence stdlib needed to be put
in the ast package. Now dumpstdlibast no longer depends on Jsonnet,
so we can get rid of this complication.

All code using ast.StdAst should now use astgen.StdAst.

### DIFF
--- a/ast/stdast.go
+++ b/ast/stdast.go
@@ -1,8 +1,0 @@
-package ast
-
-// StdAst is the AST for the standard library.
-//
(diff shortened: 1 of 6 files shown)


sparkprime:
LGTM

PR #311, opened 2019-09-01
Make parser package internal

sbarzowski:
There is no reason for external users to directly depend
on parser. It had a few random things exported as well,
namely errors and "children" functions (helpers for AST
traversal).

It was easy to extract the errors package, but I needed to leave
children in parser for now. The errors package was also
made internal, but it's a candidate for making it public
again potentially (if someone wants to display error messages
just like us). For now it's probably too incomplete anyway.

This change has a potential of breaking the existing users
since we technically remove public APIs. These were not needed
or even helpful for actually running Jsonnet code, but perhaps
someone used them anyway.


error_formatter.go
```diff
@@ -68,7 +68,7 @@ func (ef *termErrorFormatter) Format(err error) string {
 	switch err := err.(type) {
 	case RuntimeError:
 		return ef.formatRuntime(&err)
-	case parser.StaticError:
+	case errors.StaticError:
```
seh: Something to consider: Make `errors.StaticError` unexported, and have `errors.MakeStaticError` (possibly renamed to just `errors.Static`) return an `error`. Provide a function in the `errors` package like so:
```go
func SourceLocation(err error) (ast.LocationRange, bool)
```
In that function, if `err` is of type `*StaticError`, return `(err.(*StaticError)).Loc()` and `true`. You can use that here to avoid sniffing for specific error types, and could instead dispatch on what kind of information is associate with these (opaque) errors.

sbarzowski:
I have plans for some more changes of this sort (i.e. moving things around, so that it all makes more sense). Most importantly:
* Moving Children to a reasonable location (as indicated in TODO).
* Moving most of the interpreter to an internal package. Right now, it's a bit chaotic, especially in the main directory with Go files mixed with things like travis config. I hope that I can bring it closer to what seems to be considered the standard layout (i.e. something along these lines https://github.com/golang-standards/project-layout). We will still need to keep the `jsonnet` package in the root dir, so that we don't break existing imports, but I think we can limit it to a single file (namely vm.go).

error_formatter.go
sbarzowski: Hmmm... I don't see anything bad about distinguishing between error types (especially when the division is so clean as here - RuntimeError, StaticError and everything else being considered an Internal Error). IIRC the original reason for MakeStaticError was not to hide anything, but to make it harder to forget about location.

seh: That's fine. I just saw an opportunity to shrink the surface area of the `errors` package. Since it's internal, we don't want to make its types and functions easy to leak any further.

sparkprime:
LGTM

PR #313, opened 2019-09-01
Enforce no golangci-lint warnings in CI

sbarzowski:


sbarzowski:
It can use all golint warnings + quite a few others. Most importantly for CI purposes, the warnings can be silenced (and sometimes it's necessary for backward compatibility).

sbarzowski:
Of course the warnings will need to be fixed (or silenced) before merging this.

sbarzowski:
Unfortunately this change requires bumping the supported version of Go to 1.11. This is actually not that bad, since that's the oldest version officially supported by the Go project since Feb 2019.

ghostsquad:
@sbarzowski I think this can be easily merged now that minimum go version is 1.11

internal/parser/lexer.go
```diff
@@ -676,9 +673,9 @@ func (l *lexer) lexSymbol() error {
 				}
 			}
 			if allStar {
-				for _, l := range lines {
-					if l[0] == '*' {
-						l = " " + l
+				for i := range lines {
```
sbarzowski: @sparkprime I don't really understand the point of this. Why do we need this "little hack"?

FYI before my change this whole loop was a noop, because `l` was a copy.

builtins.go
```diff
@@ -464,7 +450,10 @@ func builtinSort(i *interpreter, trace traceElement, arguments []value) (value,
 		}
 	}
 
-	data.Sort()
+	err = data.Sort()
+	if err != nil {
```
sbarzowski: This is probably the most consequential part of this change. In particular one of our tests already covers this, but I have missed it before (it was committed with a bad golden).

PR #315, opened 2019-09-09
Check Bazel in Travis

### LINKED ISSUE #303: Add bazel builds to Travis
It would be good to make sure that Bazel works after every commit, without manually testing every time.

sbarzowski:
Fixes #303

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,22 @@
 language: go
 sudo: false
-go:
-  - 1.x
-  - 1.8.x
-  - tip
+matrix:
+  include:
+    - go: 1.x
+    - go: 1.8.x
+    - go: tip
+    - name: "Bazel Check"
+      go: 1.x
+      script: ./travisBazel.sh
+      before_install:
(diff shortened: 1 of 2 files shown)


sbarzowski:
Ok, it seems to be pretty much working, but I still need to figure out how to do both Bazel and regular stuff (at this point it uses Bazel exclusively).

sbarzowski:
Fixes #303

sbarzowski:
@seh @sparkprime 

Ok, I think it's pretty reasonable now. One thing I'm not sure about is whether `bazel test //:go_default_test` is the best job to run here. This seems to be limited to the `jsonnet` package (and not CLI, parser, linter or any other stuff). Perhaps we could test everything somehow? I tried using `//...` but is failed in a way I'm not able to debug.

sparkprime:
ERROR: missing input file '//ast:stdast.go'
ERROR: /home/travis/gopath/src/github.com/google/go-jsonnet/ast/BUILD.bazel:3:1: //ast:go_default_library: missing input file '//ast:stdast.go'

sbarzowski:
That's a good thing, it means that it actually catches the error that was fixed in #316. I'll rebase now and it should go away.

sparkprime:
Well then, LGTM

PR #316, opened 2019-09-09
Fix Bazel config

sbarzowski:


sbarzowski:
This fixes the issue reported by @lihaoyi-databricks: https://github.com/google/go-jsonnet/issues/303#issuecomment-529272768. The problem was trivial - I moved stuff around and forgot to run gazelle. 

I started working on https://github.com/google/go-jsonnet/pull/315 so that it does not happen again in the future.

sbarzowski:
This is a trivial sync and it's coming up in other changes when I'm running gazelle, so I'll just merge this now.

PR #317, opened 2019-09-09
Add missing cases in specialChildren

sbarzowski:
Namely:
1) function body and optional params
2) object locals


sparkprime:
LGTM

PR #318, opened 2019-09-09
Better importing (caching & public API)

sbarzowski:
* It adds new API which allows using VM for importing
  files in external tools "just as Jsonnet would". This is
  primarily intended for use in static analysis tools.
* Imports are now cached between evaluate calls. This may improve
  performance significantly for some users. I would like to add
  some way of achieving this with commandline in the future.
* Additional layer of caching was internally added - AST level.
  This was necessary so that Jsonnet could always return the same
  exact AST when asked multiple times (meaning the same pointers).

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -21,6 +21,9 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+
+	"github.com/google/go-jsonnet/ast"
+	"github.com/google/go-jsonnet/internal/program"
 )
 
 // An Importer imports data from a path.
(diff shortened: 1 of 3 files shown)


sbarzowski:
It was a side-effect of the work on the Linter which needs to go through imports to be really useful.

imports.go
```diff
@@ -97,6 +110,21 @@ func (cache *importCache) importData(importedFrom, importedPath string) (content
 	return
 }
 
+func (cache *importCache) importAST(importedFrom, importedPath string) (ast.Node, string, error) {
+	// TODO(sbarzowski) perhaps we can cache the AST, so that we go through parsing twice
```
sparkprime: This comment doesn't make sense

sparkprime:
Seems reasonable to expose this.

sbarzowski:
I fixed that obsolete comment, another typo, and a bug (`vm.flushValueCache` was empty before). 

@sparkprime Do you want to take another look at it?

sparkprime:
LGTM

PR #319, opened 2019-09-12
Add fodder to AST nodes

sparkprime:


sparkprime:
@sbarzowski There's no tests yet.  We will be able to test by using the C++ formatter tests once we have a formatter. The next step would be to write a simple pretty-printer which I plan to do.  That is equivalent to running:

```jsonnetfmt -n 0 --max-blank-lines=0 --string-style=l --commment-style=l --no-pretty-field-names --no-pad-arrays --no-pad-objects --no-sort-imports```

**Comment by sbarzowski** [comment]
> We will be able to test by using the C++ formatter tests once we have a formatter. The next step would be to write a simple pretty-printer which I plan to do.

Yeah, makes sense. If it's any easier, for the testing purposes anything which doesn't lose information will do (since we'll need to feed to the real jsonnetfmt for comparison anyway).

File: ast/ast.go [hunk 1]
```diff
@@ -53,32 +53,36 @@ type Nodes []Node
 // NodeBase holds fields common to all node types.
 type NodeBase struct {
 	LocRange LocationRange
+	Fodder   Fodder
```
sbarzowski: What is this main Fodder that goes in every node? Perhaps a comment here would be nice.

ast/ast.go
```diff
@@ -137,22 +145,37 @@ type ForSpec struct {
...
+type CommaSeparatedExpr struct {
+	Expr        Node
+	CommaFodder Fodder
 }
 
 // Arguments represents positional and named arguments to a function call
 // f(x, y, z=1).
 type Arguments struct {
-	Positional Nodes
+	Positional []CommaSeparatedExpr
```
sbarzowski: I would probably kept a separate slice of commaFodders in Arguments instead of creating CommaSeparatedExpr struct for this purpose. I think ideally Fodder business would be completely transparent to the code that does not need to care about it. 

It's not a big deal, we can leave it like this AFAIC.

ast/ast.go
```diff
@@ -335,21 +367,35 @@ type Error struct {
...
+	NameFodder  Fodder
+	Name        Identifier
+	CommaFodder Fodder
 }
 
 // Parameters represents the required positional parameters and optional named
 // parameters to a function definition.
 type Parameters struct {
-	Required Identifiers
+	Required []CommaSeparatedID
```
sbarzowski: Same as before about CommaSeparatedExpr.

ast/ast.go
sparkprime: Then if someone adds a param to the beginning or something, they would get misaligned.  I think it's better this way - at least you can copy them around opaquely without caring about the fodder.

ast/ast.go
sparkprime: Done.

PR #324, opened 2019-09-23
Disallow unary + on Object

### LINKED ISSUE #276: Disallow unary + on Object
For some reason this works here: `jsonnet -e '+ {}'`

alldroll:
This resolves #276 

In accordance with https://github.com/google/jsonnet/blob/0ecced5ba6110f3a5978a604001a0d58430f9f26/core/vm.cpp#L2901, we allow unary operation '+-~' only on numbers


sbarzowski:
Thanks!

PR #325, opened 2019-09-26
Linter

sbarzowski:


sbarzowski:
Some cleanup is still needed, but it's functional. At this point it may be worth looking at the examples and overall structure.

Later, I will provide more some explanation, rebase and fix the remaining issues (at least the important ones). 

I may also split off a few changes from it (some things were needed to get it working, but are not strictly related). I've already done that with https://github.com/google/go-jsonnet/pull/318.

sbarzowski:
I fixed the most important remaining issues and refactored it a little bit. I'll look at it a few more times, but I expect to merge it in a few days.

PR #328, opened 2019-09-29
Native callbacks in c-bindings

### LINKED ISSUE #323: Support native callbacks in c-bindings
Requirement to move Kapitan from C++ to Go

alldroll:
This resolves #323.

This adds support of native callbacks in c-bindings. I am not sure about the idea of implementation, especially with `json` handles removal policy:
* https://github.com/google/go-jsonnet/pull/328/files#diff-2cd024d05c7ef8aac7687e1e43c7f375R205
* https://github.com/google/go-jsonnet/pull/328/files#diff-2cd024d05c7ef8aac7687e1e43c7f375R353

Please give me feedback if I am not right.


sbarzowski:
Thanks. I'll take a closer look at it tomorrow or the day after that. (Feel free to ping me if I forget).

sparkprime:
It looks like you're passing the JSON tree (i.e. all of the nodes) by reference to the C++ world.  While we had to pass the VM by reference (it's a mutable thing that you need a pointer to), it's not actually necessary to pass the JSON tree by reference.  You can instead copy the Go JSON tree into a C++ heap allocated JSON tree and then pass that to the C++ world.  Thus there is no garbage collection problem as it doesn't matter if the original Go JSON tree gets garbage collected.

alldroll:
Could you please briefly describe this 
```
You can instead copy the Go JSON tree into a C++ heap allocated JSON tree and then pass that to the C++ world
```

Do you mean instead of representing `JsonnetJsonValue` as a "handle" struct and using c-bindings for `jsonnet_json_*` functions, use native c++ `jsonnet_json_*` functions and just transfer go json tree representation to c++ and backward?

sparkprime:
Yeah. I'm not sure about that, maybe Stan has some ideas. I feel like you should just be able to parse the C++ structure from the Go side by calling C++ functions?

c-bindings/c-bindings.go
```diff
@@ -4,33 +4,27 @@ import (
...
-// The way it is implemented below is simple and has low overhead, but requires us to keep
-// a list of used IDs. This results in a permanent "leak". I don't expect it to ever
-// become a problem.
-// The VM IDs start with 1, so 0 is never a valid ID and the VM's index in the array is (ID - 1).
-
-// VMs is the set of active, valid Jsonnet virtual machine handles allocated by jsonnet_make.
-var VMs = []*vm{}
-var freedIDs = []uint32{}
+type jsonValue struct {
+	obj      interface{}
```
sbarzowski: `obj` is not a very good name - objects are just one kind of JSON values. I would go with `val`.

c-bindings/c-bindings.go
```diff
@@ -4,33 +4,27 @@ import (
...
+	children []*C.struct_JsonnetJsonValue
```
sbarzowski: I would add an explicit comment that these children are exclusively owned by this `jsonValue`. Perhaps I would even change the name to `owned` (they are only used for deallocation).

c-bindings/c-bindings.go
```diff
@@ -156,5 +160,236 @@ func jsonnet_tla_code(vmRef *C.struct_JsonnetVm, key, value *C.char) {
...
+			return v.obj, nil
+		},
+	}
+
+	vm.NativeFunction(f)
+}
+
+//export jsonnet_json_extract_string
+func jsonnet_json_extract_string(vmRef *C.struct_JsonnetVm, json *C.struct_JsonnetJsonValue) *C.char {
+	_ = getVM(vmRef) // Here we check it, is it ok?
```
sbarzowski: It is ok, but we don't have to validate it at all. C++ implementation also doesn't use vm for these functions (https://github.com/google/jsonnet/blob/master/core/libjsonnet.cpp#L55).

c-bindings/c-bindings.go
```diff
@@ -156,5 +160,236 @@ func jsonnet_tla_code(vmRef *C.struct_JsonnetVm, key, value *C.char) {
...
+	table[C.GoString(f)] = val.obj
+	d.children = append(d.children, v)
+}
+
+//export jsonnet_json_destroy
+func jsonnet_json_destroy(vmRef *C.struct_JsonnetVm, v *C.struct_JsonnetJsonValue) {
+	_ = getVM(vmRef) // Here we check it, is it ok?
+
+	for _, child := range getJSONValue(v).children {
+		// TODO what if a child is already freed?
```
sbarzowski: Well, double free. But the ownership of sub-values is exclusive, so the user should not let that happen (doc: https://github.com/google/jsonnet/blob/master/include/libjsonnet.h#L133, C++ impl: https://github.com/google/jsonnet/blob/master/core/libjsonnet.cpp#L149). This is acceptable behavior to me. 

BTW I would argue that the ownership there should not be exclusive, but that boat has sailed probably, and changing that is completely out of scope in this PR. 

BTW2 we could perhaps add an explicit "cleared" state, and fail with a clear message if the user saves the same JsonnetJSONValue in two places, but I'm not sure it's worth it.

c-bindings/c-bindings.go
```diff
@@ -156,5 +160,236 @@ func jsonnet_tla_code(vmRef *C.struct_JsonnetVm, key, value *C.char) {
...
+//export jsonnet_native_callback
+func jsonnet_native_callback(vmRef *C.struct_JsonnetVm, name *C.char, cb *C.JsonnetNativeCallback, ctx unsafe.Pointer, params **C.char) {
+	vm := getVM(vmRef)
+	p := unsafe.Pointer(params)
+	sz := unsafe.Sizeof(*params)
+
+	var identifiers ast.Identifiers
+
+	for i := 0; ; i++ {
+		arg := (**C.char)(unsafe.Pointer(uintptr(p) + uintptr(i)*sz))
```
sbarzowski: `param`, not `arg`. I know it seems like a small detail, but in the context of language tools, I think it's worth it to be precise. 

Explanation: https://www.quora.com/What-is-the-difference-between-argument-and-parameters-in-C

c-bindings/c-bindings.go
```diff
@@ -156,5 +160,236 @@ func jsonnet_tla_code(vmRef *C.struct_JsonnetVm, key, value *C.char) {
 	vm.TLACode(C.GoString(key), C.GoString(value))
 }
 
+//export jsonnet_native_callback
+func jsonnet_native_callback(vmRef *C.struct_JsonnetVm, name *C.char, cb *C.JsonnetNativeCallback, ctx unsafe.Pointer, params **C.char) {
+	vm := getVM(vmRef)
+	p := unsafe.Pointer(params)
+	sz := unsafe.Sizeof(*params)
+
+	var identifiers ast.Identifiers
```
sbarzowski: `s/identifiers/paramNames/` maybe?

c-bindings/handles.go
```diff
@@ -0,0 +1,87 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+)
+
+const maxID = 1000
```
sbarzowski: I think we should raise this limit to at least `100 000` if we're using it for JSON values.

c-bindings/handles.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Because of Go GC, there are restrictions on keeping Go pointers in C.
+// We cannot just pass *jsonnet.VM/JsonValue to C. So instead we use "handle" structs in C
+// which refer to JsonnetVM/JsonnetJsonValue by a numeric id.
+// The way it is implemented below is simple and has low overhead, but requires us to keep
+// a list of used IDs. This results in a permanent "leak". I don't expect it to ever
+// become a problem.
+// The Handle IDs start with 1, so 0 is never a valid ID and the Handle's index in the array is (ID - 1).
+
+// handlesTable is the set of active, valid Jsonnet virtual machine handles allocated by jsonnet_make*.
```
sbarzowski: "virtual machine handles" - obsolete

c-bindings/handles.go
```diff
@@ -0,0 +1,87 @@
...
+// ensureValidID returns an error if the given handle ID is invalid, otherwise returns nil
+func (h *handlesTable) ensureValidID(id uint32) error {
+	if id == 0 || uint64(id) > uint64(len(h.objects)) {
+		return errInvalidHandle
+	}
+
+	// we should check freed ids too:
+	// * we should inform a client if there is an attempt to free already freed id
+	// O(len(freedIDs)) maybe need to use map?
+	for _, freed := range h.freedIDs {
```
sbarzowski: I would get rid of this check, at least for this change. In C the user needs to be careful not to free something twice and it makes the complexity of JSON operations worse (quadratic instead of linear).

If we want to catch these errors, we can do it in separate change. In particular we may consider replacing the ID accounting business with something along the lines of `map[unsafe.Pointer]interface{}`. Anyway, it's out of scope in this PR.

sparkprime:
I agree it's good enough.  A bit more complicated but pragmatic right now.

Fun fact -- the C API is missing the functions to let the client C code parse the JSON given to it in the parameter of the callback.  Why?  It never got implemented.  The main consumer of this was Python and in Python you can just use `std.toString()` on the Jsonnet side and the Python json module to parse it again on the Python side.

https://github.com/google/jsonnet/blob/master/core/vm.cpp#L2476

So indeed, we would have to extend the API to add this before we could do it the way I was proposing.  Given that passing by reference works, and I want to get Kapitan onto go-jsonnet  ASAP, we should proceed :)

ramaro:
We can now use go-jsonnet in Kapitan, thank you @alldroll!

PR #330, opened 2019-10-11
Implemented jsonnet_import_callback c-binding

### LINKED ISSUE #322: Support import callbacks in c-bindings
Requirement to move Kapitan from C++ to Go

alldroll:
This resolves #322

### DIFF
--- a/c-bindings/c-bindings.go
+++ b/c-bindings/c-bindings.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"unsafe"
 
 	"github.com/google/go-jsonnet"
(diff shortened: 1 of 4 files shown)


c-bindings/c-bindings.go
```diff
@@ -25,6 +25,32 @@ type jsonValue struct {
...
+}
+
+// Import fetches data from a given path by using c.JsonnetImportCallback
+func (i *importer) Import(importedFrom, importedPath string) (contents jsonnet.Contents, foundAt string, err error) {
+	var (
+		success   = C.int(0)
+		foundHereC *C.char
+	)
+
+	resultC := C.jsonnet_internal_execute_import(i.cb, i.ctx, C.CString(importedFrom), C.CString(importedPath), &foundHereC, &success)
```
sbarzowski: I think `C.CString(importedFrom)` is not right here. There is a slight difference in the C and Go API. The importer in Go gets the full path of the file as `importedFrom` while in C API it's just the directory it is in (the Go approach here is supposed to be better and less opinionated, but C API should be 100% compatible). Directory is understood here as *the part of the path before the last slash*. 

You can take a look at how it's handled in C++ code: https://github.com/google/jsonnet/blob/master/core/vm.cpp#L761 for how the imports are handled in C++.

c-bindings/libjsonnet.cpp
```diff
@@ -34,6 +34,22 @@ struct JsonnetJsonValue* jsonnet_internal_execute_native(JsonnetNativeCallback *
...
+                                      const char *base,
+                                      const char *rel,
+                                      char **found_here,
+                                      int *success)
+{
+    return (cb)(ctx, base, rel, found_here, success);
+}
+
+void jsonnet_internal_free_string(char *str) {
+    if (str != nullptr) {
```
sbarzowski: FYI: This `if` is not necessary. Freeing `nullptr` is a no-op (https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null). Some implementation used to be non-compliant in the past, so this check became a thing, but I don't think it's worth doing anymore.

sbarzowski:
Thank you for another great change!

c-bindings/libjsonnet.cpp
alldroll: Thank you, I'll keep it on mind!

ramaro:
We can now use go-jsonnet in Kapitan, thank you @alldroll!

PR #332, opened 2019-10-15
Fix #331 missing comma error

### LINKED ISSUE #331: missed comma should return parse error
This program:
```jsonnet
{x:1,y:1z:1}
````
is producint this output on HEAD
```
{
   "x": 1,
   "y": 1,
   "z": 1
}
```

### PR DESCRIPTION
**Pull request by sparkprime**

### CODE REVIEW

File: internal/parser/parser.go [hunk 1]
```diff
@@ -622,6 +622,8 @@ func (p *parser) parseObjectRemainder(tok *token) (ast.Node, *token, error) {
 		if next.kind == tokenComma {
 			gotComma = true
 			next = p.pop()
+		} else {
+			gotComma = false
```
seh: Wow, that variable is defined 268 lines up from here. It's easy to lose track of where it's going to be consulted and reset.

sparkprime: It'd probably benefit from splitting the cases and the ForSpec stuff out into functions, but the next token and gotComma has to be passed into all of them and some additional other variables as well in some cases

alldroll:
Maybe is it better to add some tests on this problem?

sbarzowski:
> Maybe is it better to add some tests on this problem?

A suitable example was provided by @mikedanese in https://github.com/google/go-jsonnet/issues/331 to the tests.

sparkprime:
There's a test in the test_suite (i.e. cpp github) which I haven't submitted a PR for yet.

internal/parser/parser.go
sparkprime: I was working on this but then the HDD I was doing it on died so I'll probably have to do it again :)  But it wasn't hard.  I need my new laptop to arrive first though probably.

sparkprime:
https://github.com/google/jsonnet/pull/719

internal/parser/parser.go
sparkprime: Finally re-did it #335

PR #335, opened 2019-11-05
break up parseObjectRemainder

sparkprime:


seh:
I find it much easier to follow, organized in smaller pieces like this.

PR #336, opened 2019-11-11
Use backed by go-jsonnet for python extension

### LINKED ISSUE #333: Python API
We have implemented c bindings for `_native`, `_import` callbacks.

In order to integrate `go-jsonnet` for [kapitan](https://github.com/deepmind/kapitan), we should build python API using shared library generated by `cgo` (as it implemented [here](https://github.com/google/jsonnet/blob/master/python/_jsonnet.c))

What do you think about it? Here is a slight [discussion](https://github.com/deepmind/kapitan/issues/144#issuecomment-544441385)

alldroll:
It resolves #333

I have some implementation questions:

* What is the best way to get the [version](https://github.com/google/go-jsonnet/pull/336/commits/437829ff723978ca6a9060ca7579e37f54a741d6#diff-2eeaed663bd0d25b7e608891384b7298R26)
* Should we copy entire `python` folder?

### DIFF
--- a/c-bindings/BUILD.bazel
+++ b/c-bindings/BUILD.bazel
@@ -12,6 +12,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 go_library(
     name = "go_default_library",
     srcs = [
+        "handles.go",
         "c-bindings.go",
         "internal.h",
         "json.cpp",
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ from subprocess import Popen, PIPE
 
 DIR = os.path.abspath(os.path.dirname(__file__))
 LIB_DIR = DIR + '/c-bindings'
-MODULE_SOURCES = ['cpp-jsonnet/python/_jsonnet.c']
+MODULE_SOURCES = ['python/_jsonnet.c']
 
 def get_version():
     """
@@ -46,7 +46,7 @@ class BuildJsonnetExt(BuildExt):
         BuildExt.run(self)
 
 jsonnet_ext = Extension(
-    '_jsonnet',
+    '_gojsonnet',
     sources=MODULE_SOURCES,
     extra_objects=[
         LIB_DIR + '/libgojsonnet.a',
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ MODULE_SOURCES = ['cpp-jsonnet/python/_jsonnet.c']
 
 def get_version():
     """
-    Parses the version out of libjsonnet.h
+    Parses the version out of vm.go
     """
     with open(os.path.join(DIR, 'vm.go')) as f:
         for line in f:
@@ -41,7 +41,7 @@ class BuildJsonnetExt(BuildExt):
         p.wait()
 
         if p.returncode != 0:
-            raise Exception('Could not build libgojsonnet.so')
+            raise Exception('Could not build libgojsonnet.a')
 
         BuildExt.run(self)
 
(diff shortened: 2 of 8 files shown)


c-bindings/BUILD.bazel
```diff
@@ -12,6 +12,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 go_library(
     name = "go_default_library",
     srcs = [
+        "handles.go",
```
sbarzowski: Hmmm... I wonder how it was passing tests before without that... Maybe the bazel check is not building this part. We should investigate this (ofc, out of scope for this change).

setup.py
```diff
@@ -0,0 +1,60 @@
...
+    '_jsonnet',
+    sources=MODULE_SOURCES,
+    extra_objects=[
+        LIB_DIR + '/libgojsonnet.a',
+    ],
+    include_dirs = ['cpp-jsonnet/include'],
+    language='c++',
+)
+
+setup(name='jsonnet',
```
sbarzowski: What do you think about creating a separate package `gojsonnet` backed by this implementation at first and switch over fully only when it's confirmed to be working in the wild by early adopters?

@alldroll @sparkprime

alldroll: I tend to agree with you, it is safer

sbarzowski:
Besides the fact that we don't have a decision how to handle versions here, It looks good to me. 

When we figure out how to do versions and releases here, we probably should update the release checklist to reflect that: https://github.com/google/jsonnet/blob/master/release_checklist.md

sbarzowski:
Ah, and we should run these Python API tests in CI.

alldroll:
I used the same approach for version handling as [here](https://github.com/google/jsonnet/blob/master/setup.py#L37).

setup.py
```diff
@@ -0,0 +1,69 @@
...
+from setuptools.command.build_ext import build_ext as BuildExt
+from subprocess import Popen, PIPE
+
+DIR = os.path.abspath(os.path.dirname(__file__))
+LIB_DIR = DIR + '/c-bindings'
+MODULE_SOURCES = ['cpp-jsonnet/python/_jsonnet.c']
+
+def get_version():
+    """
+    Parses the version out of libjsonnet.h
```
sbarzowski: This comment is wrong -- it's `vm.go` now.

setup.py
```diff
@@ -0,0 +1,69 @@
...
+        p = Popen(['go', 'build', '-o', 'libgojsonnet.a', '-buildmode=c-archive'], cwd=LIB_DIR, stdout=PIPE)
+        p.wait()
+
+        if p.returncode != 0:
+            raise Exception('Could not build libgojsonnet.a')
+
+        BuildExt.run(self)
+
+jsonnet_ext = Extension(
+    '_jsonnet',
```
sbarzowski: Hmmm... should we use _gojsonnet here? I'm not sure what would happen if both jsonnet and gojsonnet packages were installed and we left it like that.

alldroll:
@sbarzowski 

I have replaced ` _jsonnet` with `_gojsonnet` (it required to copy jsonnet.c to the python folder)

sbarzowski:
Thanks.

> it required to copy jsonnet.c to the python folder
That's fine. If we end up changing it a lot for whatever reason, we can extract the common parts later.

sparkprime:
This is great work and very much needed to help get people over to the Go implementation with its performance improvements and other benefits.  Thanks very much and sorry it took me so long to get to it!

sbarzowski:
I'm going to merge this now. Some minor issues may arise, e.g. compatibility with OS X and Windows may require adjustments, but we can tackle these in separate changes.

PR #342, opened 2019-12-13
feat: simplify and streamline development 

ghostsquad:
This streamlines and simplifies local development by adding
a Makefile and moving a majority of commands listed in the README to
a Make target.

Additionally,

Add support for asdf-managed versions of bazel

https://github.com/asdf-vm/asdf
https://github.com/rajatvig/asdf-bazel


ghostsquad:
@googlebot I signed it!

ghostsquad:
Not sure why the build failed with this:

```
cmd/jsonnet/cmd.go:32:2: cannot find package "github.com/fatih/color" in any of:

	/home/travis/.gimme/versions/go1.8.7.linux.amd64/src/github.com/fatih/color (from $GOROOT)

	/home/travis/gopath/src/github.com/fatih/color (from $GOPATH)

The command "./travisBuild.sh" exited with 1.
```

as none of these files were touched. Might be transient.

sbarzowski:
I have mixed feeling putting everything in Makefile and recommending that to the users - I'm not sure if we should hide these details from the users. I mean, I am definitely ok with adding a Makefile, but not so sure about changing the README. If we add the Makefile we should also verify it (at least the most important parts) with Travis.

@sparkprime Any thoughts?

ghostsquad:
The `Makefile` remains very simple. There's no "cleverness" baked in, so if you wanted to see what the commands were doing, and even run them without `make`, it's straight forward.

### Pros (If you have `make installed`)

- The commands to run are terse
- If users have autocomplete installed in some form (e.g. from Zsh) - command discovery is easy `make <tab>`
- You get all the benefits of chaining commands together for a simplified workflow

### Cons

- For those not using `make`, the commands are in the `Makefile` instead of in the `README.md`

### Other Notes

There's no good mitigation for the con listed above. If you were to give the user the option of seeing the command in the `README.md` or `Makefile`, then that leaves room for drift between what's documented, vs what's in the `Makefile`.

### Conclusion

I won't be hurt if the decision to merge this is `no`. Let me say though that using this `Makefile` as a new developer to the project has made it very easy.

ghostsquad:
I'm happy to update this PR to have Travis run these commands, which would verify the `Makefile`

ghostsquad:
@sbarzowski do you know what travis skips the go tests (`./tests.sh --skip-go-test`)?

ghostsquad:
I noticed that the commands in `./travisBuild.sh` and `./travisBazel.sh` don't match the other commands that were previously in the README that a human would run.

So, there's not much verification that would be done here without adding in arbitrary targets specific to Travis. That can certainly be done, but I don't know if it adds much value.

Please let me know how to continue with this.

sbarzowski:
> nevermind, it's because the go tests are run using bazel
No, there is a separate travis job for Bazel. They are skipped, because they are run in the same job using `  $GOPATH/bin/goveralls -service=travis-ci` - this way we get coverage reports.

> Please let me know how to continue with this.
You could add a new Travis job, which runs `make` (a new entry in `matrix.include` which overrides `script`).

Makefile
```diff
@@ -0,0 +1,60 @@
...
+# https://github.com/golang/go/issues/30515
+# We temporarily set GO111MODULE=off here to avoid adding these binaries to the go.mod|sum files
+# As they are not needed during runtime
+install.dependencies : export GO111MODULE=off
+install.dependencies:
+	git submodule init
+	git submodule update
+	go get github.com/clipperhouse/gen
+	go get github.com/clipperhouse/set
+PHONY: install.dependencies
```
sbarzowski: Shouldn't it be `.PHONY`?

ghostsquad: yes, thanks for catching.

ghostsquad: fixed.

ghostsquad:
@sbarzowski the build is currently failing on 1.8 because of this:

```
OK

cmd/jsonnet/cmd.go:32:2: cannot find package "github.com/fatih/color" in any of:
	/home/travis/.gimme/versions/go1.8.7.linux.amd64/src/github.com/fatih/color (from $GOROOT)
	/home/travis/gopath/src/github.com/fatih/color (from $GOPATH)
The command "./travisBuild.sh" exited with 1.
```

sbarzowski:
We don't want to really support go1.8 anymore, so I bumped what is considered oldest Go we support to 1.11. I expect the error to go away when you rebase.

Regarding README, I'm not sure. Could you perhaps extract README changes to a separate PR? Then we could merge this one without any further delay.

ghostsquad:
I've reverted the changes to the Readme.

sbarzowski:
@ghostsquad Could you rebase it when you have a moment? I expect the tests would pass and I would just merge it then.

ghostsquad:
@sbarzowski done

ghostsquad:
failed again. Maybe this is a caching issue?

```
cmd/jsonnet/cmd.go:32:2: cannot find package "github.com/fatih/color" in any of:

	/home/travis/.gimme/versions/go1.11.13.linux.amd64/src/github.com/fatih/color (from $GOROOT)

	/home/travis/gopath/src/github.com/fatih/color (from $GOPATH)

The command "./travisBuild.sh" exited with 1.
```

ghostsquad:
Or maybe a rate limiting problem? I don't have access to just retry that one job, but maybe it might help to run it solo.

ghostsquad:
I'm going to close this and create a new PR. Maybe there's a caching issue.

PR #343, opened 2019-12-15
feat: Improve performance of substr by 84%

ghostsquad:
feat: Improve performance of substr by ~5x

This pulls in the implementation of substr into native Go instead of interpretted Jsonnet.

```
➜ ./benchmark.sh Builtin
Running Before Test... (10s)
Running After Test... (10s)
benchmark                       old ns/op     new ns/op     delta
Benchmark_Builtin_substr-16     96131271      14533163      -84.88%
```

part of #111


builtins_benchmark_test.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	// "os"
+	"os/exec"
+	"testing"
+)
+
+func Benchmark_Builtin_substr(b *testing.B) {
+	for n := 0; n < b.N; n++ {
+		cmd := exec.Command("./jsonnet", "./builtin-benchmarks/substr.jsonnet")
+		// cmd.Stdout = os.Stdout
```
sbarzowski: Have you left these lines commented out intentionally?

builtins.go
```diff
@@ -1211,6 +1264,7 @@ func buildBuiltinMap(builtins []builtin) map[string]evalCallable {
 
 var funcBuiltins = buildBuiltinMap([]builtin{
 	builtinID,
+	&ternaryBuiltin{name: "substr", function: builtinSubstr, parameters: ast.Identifiers{"str", "from", "len"}},
```
sbarzowski: Could you put it before `splitLimit`  (both the table entry and the implementation)? While the list follows no strict order, it's good to have similar functions together, e.g. string functions next to other strings functions, math functions next to other math functions etc.

ghostsquad: done

builtins_benchmark_test.go
ghostsquad: yes, they are useful for debugging. In the future I might decide on a better way.

I'm not even sure if this is the right way to do this testing. If you have any suggestions, I'm open! :)

sbarzowski: I think it's ok for now as a first step. I was running `jsonnet` command by hand for this purpose. 

If we're going to use Go benchmarks for that, we probably should allow passing the path to Jsonnet command by commandline. The output redirection thing could also be a flag,

builtins.go
```diff
@@ -839,6 +839,59 @@ func builtinPow(i *interpreter, trace traceElement, basev value, expv value) (va
 	return makeDoubleCheck(i, trace, math.Pow(base.value, exp.value))
 }
 
+func builtinSubstr(i *interpreter, trace traceElement, inputStr, inputFrom, inputLen value) (value, error) {
+	strV, ok := inputStr.(*valueString)
+	if !ok {
+		msg := fmt.Sprintf("substr first parameter should be a string, got %T", inputStr)
```
sbarzowski: I've just noticed that it will print the Go name for the type in the error message, e.g. "got *jsonnet.valueNumber". We should use Jsonnet names for that: `inputStr.getType().name`. There exist some convenience functions for checking types, such as `i.getNumber`. The errors they produce are not as nice as we have here - they don't include the information which parameter is affected (but perhaps that can be fixed, instead of doing it manually? - probably out of scope for this PR).

builtins_benchmark_test.go
ghostsquad: I agree. The way I did it worked, but was tedious.

1. comment out the line in the `builtins.go` file that overwrote the upstream jsonnet implementation
2. build the jsonnnet binary
3. run the benchmark
4. uncomment the line in `builtins.go`
5. build the binary
6. run the benchmark

builtins.go
ghostsquad: I'll fix the type part here, but I explicitly used these errors because they match the errors seen in upstream jsonnet stdlib. I thought it would be nice to keep that compatibility.

I'll write some unit tests for this as well.

builtins_benchmark_test.go
ghostsquad: I agree. The way I did it worked, but was tedious.

1. comment out the line in the `builtins.go` file that overwrote the upstream jsonnet implementation
2. build the jsonnnet binary
3. run the benchmark
4. uncomment the line in `builtins.go`
5. build the binary
6. run the benchmark

builtins.go
ghostsquad: @sbarzowski do you have an example of a unit test for these builtins?

benchmark.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
```
sbarzowski: This should probably auto-switch the directory and/or check if it's in the right one, especially since we're rm -rf-ing stuff

benchmark.sh
```diff
@@ -0,0 +1,13 @@
...
+FILTER=${1:-Builtin}
+
+mkdir -p ./builtin-benchmark-results
+rm -rf ./builtin-benchmark-results/*
```
sbarzowski: -r is actually not necessary and it makes it more dangerous

builtins.go
```diff
@@ -839,6 +839,59 @@ func builtinPow(i *interpreter, trace traceElement, basev value, expv value) (va
...
+		return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+	}
+
+	lenV, err := i.getNumber(inputLen, trace)
+	if err != nil {
+		msg := fmt.Sprintf("substr third parameter should be a number, got %T", inputLen.getType().name)
+		return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+	}
+
+	if math.Mod(lenV.value, 1) != 0 {
```
sbarzowski: This could be done slightly more elegantly IMO, with something like math.Trunc(lenV.value) == lenV.value.  Same with fromV before.

builtins.go
sbarzowski: It's enough to add an appropriately named `.jsonnet` file to the `testdata/` directory and run `go test -update` to generate output for it (fyi it regenerates output for all the tests).

sbarzowski: BTW `inputFrom.getType().name` is a string, so you need to use `%s`, not `%T` in the format string in these conditions.

benchmark.sh
sbarzowski: Something like that could help (changes directory to the script location):
```
cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
```

ghostsquad:
Thank you for the great feedback. I'll get these changes in.

builtins.go
ghostsquad: > It's enough to add an appropriately named `.jsonnet` file to the `testdata/` directory and run `go test -update` to generate output for it (fyi it regenerates output for all the tests).

Does this only test the "happy" paths? How might I test the "sad" paths?

ghostsquad: @sbarzowski I see several naming conventions for file names in `testdata` including:

`builtin_<name><#>`
`builtin<Name>_<shortDescription>`
`builtin<Name><ShortDescription>`
`<name>`
`<name>_<short>_<description>`
`std.<name><#>`
`std.<name>_<short>_<description>`

Do you have a preference for file naming here?

benchmark.sh
ghostsquad: fixed

builtins.go
ghostsquad: updated

benchmark.sh
ghostsquad: fixed

builtins.go
sbarzowski: No strong preference. It's not something I gave much thought so far. 
If they are related to a particular function it's generally it's good to include the verbatim name, with the same capitalization so that it's greppable. I tend to use description instead of numbers these days. 

(The next step for tests organization would be allowing subdirectories and having some high-level categorization. After that we may consider introducing some explicit naming convention for the tests.)

Issue #341, opened 2019-12-02
base64 stack overflow go-jsonnet


bernadinm:
# Issue std.base64 with go-jsonnet

There is a max file limit with [go-jsonnet](https://github.com/google/go-jsonnet) where if you attempt to base64 anything greater than the file size of 51695 * 2 = `103390`, you will get a stack overflow. This issue is not present with [jsonnet](https://github.com/google/jsonnet) c++ version.

```
# Randomly generate a file
openssl rand -hex 51696 > prebase64.txt # Writes double the size of the hex value + 1
```

### go-jsonnet output
```
go get github.com/google/go-jsonnet/cmd/jsonnet
$GOPATH/bin/jsonnet -e 'local test = importstr "./prebase64.txt"; std.base64(test)' > out && cut -c -5 out
```

#### Output Fail
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow

runtime stack:
runtime.throw(0x119566a, 0xe)
	/usr/local/opt/go/libexec/src/runtime/panic.go:617 +0x72
runtime.newstack()
	/usr/local/opt/go/libexec/src/runtime/stack.go:1041 +0x6f0
runtime.morestack()
	/usr/local/opt/go/libexec/src/runtime/asm_amd64.s:429 +0x8f

goroutine 1 [running]:
github.com/google/go-jsonnet.(*interpreter).evaluate(0xc00009a0a0, 0x11efb40, 0x13538e0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/user-id/.go/src/github.com/google/go-jsonnet/interpreter.go:275 +0x41bb fp=0xc024000a80 sp=0xc024000a78 pc=0x112488b
github.com/google/go-jsonnet.(*interpreter).evaluate(0xc00009a0a0, 0x11ef800, 0x1403740, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/user-id/.go/src/github.com/google/go-jsonnet/interpreter.go:430 +0x683 fp=0xc024001338 sp=0xc024000a80 pc=0x1120d53
github.com/google/go-jsonnet.(*interpreter).evaluate(0xc00009a0a0, 0x11ef480, 0x14462c0, 0x1, 0x1, 0xc001872a20, 0x0, 0x0)
[...]

PR #344, opened 2019-12-15
feat: improve stdlib base64 performance by 98%+

ghostsquad:
```
benchmark                                 old ns/op       new ns/op     delta
Benchmark_Builtin_base64-16               10805730974     23158636      -99.79%
Benchmark_Builtin_base64_bytearray-16     8682808704      123360964     -98.58%
```

### Notes

~This PR contains commits from another pending PR. I'll rebased once #343 is merged.~


ghostsquad:
resolves #341

builtins.go
```diff
@@ -664,6 +665,16 @@ func builtinMd5(i *interpreter, trace traceElement, x value) (value, error) {
 	return makeValueString(hex.EncodeToString(hash[:])), nil
 }
 
+func builtinBase64(i *interpreter, trace traceElement, input value) (value, error) {
+	str, err := i.getString(input, trace)
```
sbarzowski: This should also work with arrays  of bytes, see: https://github.com/google/jsonnet/blob/552d8ec6f6b973a6357b83eb9bacd707366d28f0/stdlib/std.jsonnet#L1072

The string version should error out when there are characters with codepoints > 255 and those <= should use codepoints-as-bytes encoding (the current implementation results in UTF8 encoding, which would be a better option, if we didn't need to worry about backwards compatibility). See: https://github.com/google/jsonnet/blob/552d8ec6f6b973a6357b83eb9bacd707366d28f0/stdlib/std.jsonnet#L1108

And we should have tests to cover all that :-).

ghostsquad: Thanks for the feedback. I'll add that functionality and add tests as well.

ghostsquad: done

ghostsquad: implemented for arrays and strings

ghostsquad:
@sbarzowski this has been rebased and is ready for review/merge

builtins.go
```diff
@@ -664,6 +665,63 @@ func builtinMd5(i *interpreter, trace traceElement, x value) (value, error) {
...
+		if !(v < 256 && v >= 0) {
+			msg := fmt.Sprintf("base64 encountered invalid codepoint value in the array (must be 0 <= X <= 255), got %d", v)
+			return msg, false
+		}
+
+		return "", true
+	}
+
+	vStr, err := i.getString(input, trace)
+	if err == nil {
```
sbarzowski: I don't like the assumption that any error returned by `i.getString` means simply that input was not a string. This function returns a generic error which may be anything in principle. If another error condition is currently possible or will be possible in the future versions, we'll have a really confusing bug here.

One more elegant approach would be to use a type switch here. Example: https://github.com/google/go-jsonnet/blob/v0.14.0/builtins.go#L44.

builtins.go
```diff
@@ -664,6 +665,63 @@ func builtinMd5(i *interpreter, trace traceElement, x value) (value, error) {
 	return makeValueString(hex.EncodeToString(hash[:])), nil
 }
 
+func builtinBase64(i *interpreter, trace traceElement, input value) (value, error) {
+	var byteArr []byte
+
+	var sanityCheck = func(v int) (string, bool) {
+		if !(v < 256 && v >= 0) {
```
sbarzowski: A very minor nit: I think a more idiomatic way of checking if value is in the range would be:
`0 <= v && v < 256`. It's the closest you can get in most programming languages to the mathy notation 0 <= v < 256. 

Alternatively eliminating the negation would also make it more clear: `v < 0 || v > 256` than the current expresssion.

builtins.go
ghostsquad: that looks good. I'll update

builtins.go
ghostsquad: I like the `||` variant, might be written as `0 > v || v > 256`

ghostsquad: no.. that is a little confusing to look at. I'm going to use `v < 0 || v > 255`

builtins.go
```diff
@@ -681,6 +682,69 @@ func builtinMd5(i *interpreter, trace traceElement, x value) (value, error) {
...
+		if v < 0 || v > 256 {
+			msg := fmt.Sprintf("base64 encountered invalid codepoint value in the array (must be 0 <= X <= 255), got %d", v)
+			return msg, false
+		}
+
+		return "", true
+	}
+
+	switch inputType := input.(type) {
+	case *valueString:
```
sbarzowski: valueString is an interface type these days, so no * here.

ghostsquad: fixed

builtins.go
```diff
@@ -681,6 +682,69 @@ func builtinMd5(i *interpreter, trace traceElement, x value) (value, error) {
 	return makeValueString(hex.EncodeToString(hash[:])), nil
 }
 
+func builtinBase64(i *interpreter, trace traceElement, input value) (value, error) {
+	var byteArr []byte
+
+	var sanityCheck = func(v int) (string, bool) {
+		if v < 0 || v > 256 {
```
cbandy: This was originally `!(v < 256 && v >= 0)`. If I understand correctly, the same logic should be `v >= 256` or `v > 255` here.

One more bike shed... 😝 I prefer the mathy/range look of less-than: `v < 0 || 255 < v`

ghostsquad: oh yes.. the number should be 255.

ghostsquad: I agree that `N < v && v < M` looks best. However, it reads differently when it's an `||` (or). And I think `v < N || v > M` feels better.

ghostsquad: wait.. I now see the desired range in `v < 0 || 255 < v`...

PR #346, opened 2019-12-15
feat: improve std.base64Decode performance 97%+

ghostsquad:
Provides a Go-native implementation of std.base64Decode which improves performance by ~400x

Windows Results:
before: Benchmark_Builtin_base64Decode-16    1  21700901900 ns/op
after:  Benchmark_Builtin_base64Decode-16  220     53261991 ns/op

related to #111

### Notes

This includes commits from another pending PR (#343) and will be rebased once that PR is merged.


builtins.go
```diff
@@ -898,6 +952,20 @@ func builtinStrReplace(i *interpreter, trace traceElement, strv, fromv, tov valu
...
+	if err != nil {
+		return nil, err
+	}
+
+	sDec, err := base64.StdEncoding.DecodeString(string(str.value))
+	if err != nil {
+		return nil,  i.Error(fmt.Sprintf("failed to decode string: %s", err), trace)
+	}
+
+	return makeValueString(string(sDec)), nil
```
sbarzowski: Here you implictly assume UTF-8 encoding. Sadly, for compatibility reasons we need to use "naive" encoding instead (codepoint = byte, all codepoints must be <=255). Because of that this function is pretty much deprecated in favor of `std.base64DecodeBytes` which requires an explicit choice of string encoding.

ghostsquad: ok. I'll keep working on base64DecodeBytes, and I'll update this to use that method once it's ready.

ghostsquad:
@sbarzowski this PR has been rebased and is ready for review/merging

builtins.go
```diff
@@ -951,6 +952,73 @@ func builtinStrReplace(i *interpreter, trace traceElement, strv, fromv, tov valu
...
+
+	strLen := len(str)
+	if strLen%4 != 0 {
+		msg := fmt.Sprintf("input string appears not to be a base64 encoded string. Wrong length found (%d)", strLen)
+		return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+	}
+
+	for _, r := range []rune(str) {
+		n := int(r)
+		msg, ok := sanityCheck(n)
```
sbarzowski: Not sure if this check is required on this side. I think anything outside of the base64 alphabet (which contains only 64 characters from ASCII) is going to cause an error with 
`base64.StdEncoding.DecodeString(str)` - a much stricter condition than this (https://play.golang.org/p/H3_00J8w7RM).

ghostsquad: ok

builtins.go
```diff
@@ -968,6 +969,56 @@ func builtinStrReplace(i *interpreter, trace traceElement, strv, fromv, tov valu
...
+func builtinBase64DecodeBytes(i *interpreter, trace traceElement, input value) (value, error) {
+	vStr, err := i.getString(input, trace)
+	if err != nil {
+		msg := fmt.Sprintf("base64DecodeBytes requires a string, got %s", input.getType().name)
+		return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+	}
+
+	decodedBytes, err := base64DecodeGoBytes(i, trace, vStr.getGoString())
+	if err != nil {
+		return nil, i.Error(fmt.Sprintf("failed to base64 decode: %v", err.Error()), trace)
```
sbarzowski: This results in "RUNTIME ERROR: failed to base64 decode: RUNTIME ERROR: something something". Wrapping one runtime error in another is not ideal.  It's better to just pass through err here.

builtins.go
```diff
@@ -968,6 +969,56 @@ func builtinStrReplace(i *interpreter, trace traceElement, strv, fromv, tov valu
...
+func builtinBase64Decode(i *interpreter, trace traceElement, input value) (value, error) {
+	vStr, err := i.getString(input, trace)
+	if err != nil {
+		msg := fmt.Sprintf("base64DecodeBytes requires a string, got %s", input.getType().name)
+		return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+	}
+
+	decodedBytes, err := base64DecodeGoBytes(i, trace, vStr.getGoString())
+	if err != nil {
+		return nil, i.Error(fmt.Sprintf("failed to base64 decode: %v", err.Error()), trace)
```
sbarzowski: Same as above, this results in "RUNTIME ERROR ...: RUNTIME ERROR ..." and passing err through should suffice.

ghostsquad:
Ok. I'll do that. I think my next PR will be to revamp how errors work, and
stack traces. I think that's a major problem right now.

Cheers,
Wes

builtins.go
ghostsquad: fixed

builtins.go
ghostsquad: fixed

sbarzowski:
> Ok. I'll do that. I think my next PR will be to revamp how errors work, and
stack traces. I think that's a major problem right now.

Great. There's definitely space for improvement there. I'll be happy to discuss that :-).

PR #350, opened 2020-01-05
Feature/local development improvements

ghostsquad:
This streamlines and simplifies local development by adding
a Makefile based on the commands in the README.md

Add support for asdf-managed versions of bazel

Add Make Check to Travis

Supersedes #342


ghostsquad:
@sbarzowski I added a couple more Go version targets to travis.yml to try to narrow down the problem. It looks like this is only happening in go < 1.13

I'm not super familiar with go modules, but I did find this: https://github.com/fatih/color/blob/master/go.mod

```
module github.com/fatih/color

go 1.13

require (
	github.com/mattn/go-colorable v0.0.9
	github.com/mattn/go-isatty v0.0.3
	golang.org/x/sys v0.0.0-20180202135801-37707fdb30a5 // indirect
)
```

possibly indicating that the minimum requirement for `fatih/color` is 1.13?

ghostsquad:
@sbarzowski I think this is ready for re-review. I was able to get the build working.

PR #356, opened 2020-01-22
Fix build instructions and git clone

toqueteos:
It seems the build instructions were outdated and/or incorrect.

These are the changes I had to make in order to build the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,13 @@ go get github.com/google/go-jsonnet/cmd/jsonnet
 ## Build instructions (go 1.11+)
 
 ```bash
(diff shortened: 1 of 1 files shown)


sbarzowski:
Many thanks for finding and reporting that! The clone command was definitely wrong.

Can you check again that there is a problem with `go build ./cmd/jsonnet`? It seems to work just right for me.

toqueteos:
@sbarzowski You are correct, I ran `go build cmd/jsonnet` which fails because it's not looking in the local path. One moment I'll send a commit reverting that part.

sbarzowski:
Thanks a lot!

PR #363, opened 2020-02-01
Run tests on other CPU architectures (s390x, arm64, ppc64le).

sbarzowski:


sbarzowski:
A modified version of this (without s390x) was merged as part of #364. Closing this.

Issue #371, opened 2020-02-09
Cannot install gojsonnet from PyPI (experimental Python support)


sbarzowski:
```
❯ sudo pip install gojsonnet
Collecting gojsonnet
  Downloading https://files.pythonhosted.org/packages/b3/f9/58c1657578f3a8d47b73effdc856a7ce6841711298f22dee61faaf05ff57/gojsonnet-0.15.0.tar.gz
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-4herb7e2/gojsonnet/setup.py'"'"'; __file__='"'"'/tmp/pip-install-4herb7e2/gojsonnet/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-4herb7e2/gojsonnet/pip-egg-info
         cwd: /tmp/pip-install-4herb7e2/gojsonnet/
    Complete output (7 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-4herb7e2/gojsonnet/setup.py", line 63, in <module>
        version=get_version(),
      File "/tmp/pip-install-4herb7e2/gojsonnet/setup.py", line 29, in get_version
        with open(os.path.join(DIR, 'vm.go')) as f:
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-install-4herb7e2/gojsonnet/vm.go'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

iwoloschin:
A very quick glance at this makes me think you might just be not including `vm.go` in the python package.  There's a few ways to do that, the easiest is probably to specify `MANIFEST.in` (https://docs.python.org/3/distutils/sourcedist.html#the-manifest-in-template).  Alternatively, you may also be able to use `setuptools-scm` (https://pypi.org/project/setuptools-scm/) to just get version information directly from git.

I can probably take a stab at this if you'd like.

sbarzowski:
> I can probably take a stab at this if you'd like.

That would be great!

[...]

PR #373, opened 2020-02-11
Add necessary files to MANIFEST.in

iwoloschin:
Here's a quick fix for #371 .  I've done the bare minimum required to create a sdist (or wheel if you have a way to build them) that can be pip installed.  To test:

```
python3 setup.py sdist
pip install dist/gojsonnet-0.15.0.tar.gz
```

You should then be able to `import _gojsonnet` in the Python REPL.

Fair warning, I have no idea why or how the c-bindings are finding `libjsonnet.h`, I mostly just followed the error messages and included the missing files until it worked.  If there's a better way to include that I'm happy to make changes.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+include vm.go
+include cpp-jsonnet/include/libjsonnet.h
+graft c-bindings


iwoloschin:
@googlebot I signed it!

iwoloschin:
It just occurred to me that grafting c-bindings may not be ideal, if you've already run the `go build ...` command you may also package build artifacts.  It's fine if you're building on a clean repo, but it may make more sense to just switch to including each necessary file in c-bindings?  I assume this set of files does not change often, so updating `MANIFEST.in` wouldn't be too much of a burden?

sbarzowski:
Hmmm... but if it needs the Go toolchain, it means that it is actually compiling stuff. And if it's actually compiling stuff then it probably needs the whole go-jsonnet (otherwise vm.go wouldn't compile). So we need a lot more stuff for the source distribution to work.

Re build artifacts - I think we can take the same approach as go-jsonnet and use globbing to include files with the right extensions, without having to specify every single one: https://github.com/google/jsonnet/blob/master/MANIFEST.in

BTW it seems that with binary distribution it just works - I was able to build, install and import a package built with `python setup.py bdist_wheel` and I'll push that to PyPI later. It would still be nice if we fixed the MANIFEST.in here, so that the source distribution also works.

iwoloschin:
Go is required for this line:

https://github.com/google/go-jsonnet/blob/master/setup.py#L40

sbarzowski:
FYI I tried uploading the binary distribution, but unfortunately that's not as trivial as I thought. PyPI has special requirements - https://www.python.org/dev/peps/pep-0513/ (intended to ensure compatibility across distros) and doesn't accept wheels which were just built with `python setup.py bdist_wheel`. I will need to look more into this.

PR #379, opened 2020-02-18
Compile package in 32-bit mode

### LINKED ISSUE #378: Add tests for 32-bit to Travis
As far as I can tell, there is no i686 arch in Travis available, but we can just compile stuff in 32-bit mode.

It should let us catch the problems like #376 early.

alldroll:
This resolves #378

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,14 @@ matrix:
       arch: arm64
       env:
         - PYTHON_COMMAND=python3
+    - name: "arch: i686"
+      go: 1.13.x
+      arch: amd64
+      env:
+        - PYTHON_COMMAND=python3
+        - GOARCH=386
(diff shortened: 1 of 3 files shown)


sbarzowski:
Great, thanks! I'll merge this now, but I have a question. Why do we need to skip python bindings tests for i686?

alldroll:
> Great, thanks! I'll merge this now, but I have a question. Why do we need to skip python bindings tests for i686?

As I understand, there is no possibility to load a 32-bit compiled shared library on a 64-bit machine. (python bindings tests load shared library for testing)

Please correct me if I am not right.

sbarzowski:
Right, we would need to have 32-bit Python to do that (which is definitely out of scope here, and probably not worth it anyway).

PR #380, opened 2020-02-18
Replace `setup.py test` with pytest command

### LINKED ISSUE #369: `setup.py test` is deprecated
Part of tests.sh output:

```
+ python setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

We should use something else then, I guess.

alldroll:
This resolves #369 

I decided to use `pytest` as a testing framework. It has some restrictions with python versions and I am not sure it is a good choice. Please let me know if my concerns are fair.

p.s. I manually disabled `setup.py test` command.

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -14,8 +14,8 @@ then
 else
     c-bindings-tests/run.sh
 
-    $PYTHON_COMMAND setup.py build
-    pytest python
+    $PYTHON_COMMAND setup.py build --build-platlib .
+    $PYTHON_COMMAND -m pytest python
 fi
 
 export IMPLEMENTATION=golang
--- a/tests.sh
+++ b/tests.sh
@@ -15,7 +15,7 @@ else
     c-bindings-tests/run.sh
 
     $PYTHON_COMMAND setup.py build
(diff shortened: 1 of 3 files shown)


sbarzowski:
Well, the tests are failing with it, that's a problem for sure :-). Generally, using pytest seems reasonable to me.

tests.sh
```diff
@@ -14,8 +14,8 @@ then
 else
     c-bindings-tests/run.sh
 
-    $PYTHON_COMMAND setup.py build
-    $PYTHON_COMMAND setup.py test
+    $PYTHON_COMMAND setup.py build --build-platlib .
```
sbarzowski: Can you add a comment explaining why the `--build-platlib` option is needed?

alldroll: If we run `setup.py build`, it will build the package underneath 'build/'. `--build-platform` allows us to build the required `shared library object` to the current folder in order to run `pytest` without this error: `ImportError: No module named '_gojsonnet'`

PR #383, opened 2020-02-23
Avoid float overflow errors at parse time.

sparkprime:


sparkprime:
Seems like the failing test environments are due to something weird, nothing to do with this.

sbarzowski:
FYI I am a bit worried about the performance implications of this - not every time a number is used we parse it from string. It's fine now and it unblocks important work, but we may need to go back to this.

sparkprime:
My thinking was that since we also do a memory allocation that will need garbage collecting at some point, the float parsing isn't a big deal.

Issue #374, opened 2020-02-13
Python bindings: "importer problem: a different instance of Contents returned"


ramaro:
When trying out the python bindings (commit 31b9ace0f65a5a5717ed9521b57d8d9e18f24070) , I stumbled upon this error:

```
$ py main.py 
Traceback (most recent call last):
  File "main.py", line 33, in <module>
    print(jsonnet_eval("one.jsonnet"))
  File "main.py", line 29, in jsonnet_eval
    return _gojsonnet.evaluate_file(file, import_callback=_search_imports)
RuntimeError: INTERNAL ERROR: (CRASH) importer problem: a different instance of Contents returned when importing "dir/zero.jsonnet" again
goroutine 17 [running, locked to thread]:
runtime/debug.Stack(0xc0002ec910, 0x7f9a35be24a0, 0xc000099330)
```

Here's how to reproduce:

```
$ tree 
.
├── dir
│   └── zero.jsonnet
├── main.py
├── one.jsonnet
└── two.jsonnet
```

```jsonnet
// dir/zero.jsonnet 
{
  zero: 0
}
```

```jsonnet
// one.jsonnet
local two = import "two.jsonnet";
local zero = import "zero.jsonnet";

{
  one: 1,
} + two + zero
```

```jsonnet
// two.jsonnet 
local zero = import "zero.jsonnet";

{
  two: 2
} +  zero
```

Using gojsonnet v0.15.0 (commit 31b9ace0f65a5a5717ed9521b57d8d9e18f24070) works as expected:

``` bash
$ jsonnet one.jsonnet -J . -J dir
{
   "one": 1,
   "two": 2,
   "zero": 0
}
```

However, the following python snippet using import callbacks errors with error reported at the top:

```python
# main.py
import os
import _gojsonnet


def read_import(file_path):
    with open(file_path) as f:
        content = f.read()
        return file_path, content


def search_imports(cwd, import_str, search_paths):
    import_path = os.path.join(cwd, import_str)
    if os.path.exists(import_path):
        try:
            return read_import(import_path)
        except:
            raise
    else:
        for path in search_paths:
            import_path = os.path.join(path, import_str)
            if os.path.exists(import_path):
                return read_import(import_path)
    raise RuntimeError(f"Could not find {import_str}")


def jsonnet_eval(file):
    def _search_imports(cwd, import_str):
        return search_imports(cwd, import_str, [".", "dir", ])
[...]

PR #384, opened 2020-02-29
[Bindings] Fix compatibility issue with C import callback api

sbarzowski:


sbarzowski:
Fixes #374. It took me a while to get back to it, but it should work now. 

@ramaro

PR #387, opened 2020-03-03
Positional params after optional ones

### LINKED ISSUE #314: Positional argument after a named argument is not allowed
```bash
$ cat foo.jsonnet
local f(x, y = 123, z) = x + y + z;
f(1, 2, 3)

$ jsonnet/bazel-bin/cmd/jsonnet foo.jsonnet
6

$ ./sjsonnet.jar foo.jsonnet
6

$ go-jsonnet/jsonnet  foo.jsonnet
foo.jsonnet:1:21-22 Positional argument after a named argument is not allowed

local f(x, y = 123, z) = x + y + z;
```

This seems to be a new error in the go-jsonnet implementation. The old C++ google/jsonnet or Scala databricks/sjsonnet don't have any problems with no-default-value params like `z` following with-default-value params like `y`. Anyway, the message doesn't make much sense because whether an argument as a default or not, it can always be passed either positionally or via its name.

This is on the following version:

```bash
$ go-jsonnet/jsonnet
Jsonnet commandline interpreter v0.13.0
```

sparkprime:
Fixes #314


sparkprime:
I still need to update the spec.  This unblocks the Go formatter though.

internal/parser/parser.go
```diff
@@ -210,33 +209,68 @@ func (p *parser) parseArguments(elementKind string) (*token, *ast.Arguments, boo
...
+			// gotComma can be true or false here.
+			parenR = p.pop()
+			break
+		}
+
+		if !first && !gotComma {
+			return nil, nil, false, errors.MakeStaticError(fmt.Sprintf("Expected a comma before next %s, got %s.", elementKind, next), next.loc)
+		}
+
+		idFodder, id, eqFodder, expr, err := p.parseArgument()
```
sbarzowski: Maybe we can have `parseParameter` now? (See also the comment below.)

internal/parser/parser.go
```diff
@@ -210,33 +209,68 @@ func (p *parser) parseArguments(elementKind string) (*token, *ast.Arguments, boo
...
+			comma := p.pop()
+			gotComma = true
+			commaFodder = comma.fodder
+		} else {
+			gotComma = false
+		}
+
+		if id == nil {
+			// Param has no default, is just 'id2'
+			idFodder2, id2, ok := astVarToIdentifier(expr)
```
sbarzowski: Since we have separated this from `parseArguments` anyway, maybe we can get rid of this hack? I mean, let's create a new function `parseParameter` which treats a single identifier as parameter name, not as a variable reference. I think we'll be able to make the code here much nicer then (e.g. we won't need `id == nil` condition at all, we'll be able to refactor parseArgument/parseParameter to return the argument/parameter struct instead of its composite parts).

sbarzowski:
Nice!

PR #393, opened 2020-03-19
Remove bindings debug msg

ramaro:
Removing a rather noisy debug message in c-bindings which I believe was added for that purpose only and should have been deleted.


sbarzowski:
Thanks! I forgot to remove it.

Of course we should never print anything to stdout/stderr when the library interface is used.

PR #395, opened 2020-03-20
feat(api): expose formatter

sh0rez:
Exposes what I consider a good public API from the `internal/formatter`
package.

Having a go-native formatter is an awesome thing, especially because it
allows shipping the formatter as part of other projects. It's also fair
to keep the formatter's code in `internal`, yet third party projects
should have a way to use relevant parts of it.

**Note:** Ran `bazel run //:gazelle`, it modified other formatter related `BUILD.bazel` files as well ... is this expected?

### DIFF
--- a/cmd/jsonnetfmt/BUILD.bazel
+++ b/cmd/jsonnetfmt/BUILD.bazel
@@ -7,7 +7,7 @@ go_library(
     visibility = ["//visibility:private"],
     deps = [
         "//:go_default_library",
-        "//cmd:go_default_library",
+        "//cmd/internal/cmd:go_default_library",
         "//internal/formatter:go_default_library",
         "@com_github_fatih_color//:go_default_library",
     ],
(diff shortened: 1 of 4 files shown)


sh0rez:
Any chance to have this merged? @sbarzowski

sbarzowski:
Ok, but it's quite likely that we will change this interface a little one day. For example we may want to version the formatting stuff separately from the interpreter, which would mean moving that to a separate repo. Cutting some options is also a possibility (`gofmt`-style). Generally, I am not sure in which direction it should go.

sh0rez:
> Ok, but it's quite likely that we will change this interface a little one day. For example we may want to version the formatting stuff separately from the interpreter, which would mean moving that to a separate repo. Cutting some options is also a possibility (`gofmt`-style). Generally, I am not sure in which direction it should go.

That's totally fair, both things you mentioned are nice to have. For now, this got us covered to ship the formatter. Thanks!

PR #399, opened 2020-04-02
fix order-dependent test

mikedanese:
TestSdumpLargeDefinition is depnding on a global state change made by TestSdumpReusedPointers. It can't pass if it runs before TestSdumpReusedPointers.

Fixes internal auto-filled bug.


sbarzowski:
Thanks!

PR #412, opened 2020-05-21
Fix duplicated fodder (#411).

### LINKED ISSUE #411: jsonnetfmt edge case
We ran into an edge case where `go-jsonnet`'s jsonnetfmt doesn't behave as in the c++ version.

Minimal test case:

```
foo(
  bar()
  ,
  baz=qux
)
```

go-jsonnet changes this to:

```
foo(
  bar()
  ,
  baz=qux
  
)
```

This creates issues for several reasons:

* It adds trailing spaces which get deleted by many editors.
* It is not idempotent, re-running `jsonnetfmt -i test.jsonnet` will cause lines to be added indefinitely.

We worked around it by changing the original code to:

```
foo(
  bar(),
  baz=qux
)
```

sbarzowski:
CommaFodder was not cleared between arguments,
so when there was no commaFodder and the argument
was named a previous one was duplicated.

Fixes #411.

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -146,10 +146,10 @@ func (p *parser) parseArgument() (ast.Fodder, *ast.Identifier, ast.Fodder, ast.N
 func (p *parser) parseArguments(elementKind string) (*token, *ast.Arguments, bool, errors.StaticError) {
 	args := &ast.Arguments{}
 	gotComma := false
-	commaFodder := ast.Fodder{}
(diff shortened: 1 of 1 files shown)


sbarzowski:
Eh, the test don't pass, because the linter failed to install. I hope it will go away later. 

(BTW I think we should consider trying github actions instead of Travis, maybe we'll be able to have a more stable environment then. Also maybe I can pin a specific version of the linter and it's deps.)

PR #421, opened 2020-06-10
Throw an error on negative shifts

### LINKED ISSUE #416: Throw error on negative shifts
Currently it most likely will return 0.

alldroll:
It resolves #416


builtins.go
```diff
@@ -866,9 +866,35 @@ func liftBitwise(f func(int64, int64) int64) func(*interpreter, traceElement, va
...
+		}
+		y, err := i.getNumber(yv, trace)
+		if err != nil {
+			return nil, err
+		}
+		if x.value < math.MinInt64 || x.value > math.MaxInt64 {
+			msg := fmt.Sprintf("Bitwise operator argument %v outside of range [%v, %v]", x.value, int64(math.MinInt64), int64(math.MaxInt64))
+			return nil, makeRuntimeError(msg, i.getCurrentStackTrace(trace))
+		}
+		if y.value < 0 {
```
sbarzowski, line 883: Have you considered just adding this condition to the old implementation using liftBitwise? I think that would be shorter.

alldroll: My bad. Thanks:) Please take a look, is it ok?

sbarzowski: Ah, I see now. it wasn't possible to add the condition here: https://github.com/google/go-jsonnet/pull/421/files#diff-c9e72ec76c1a7488aaf4cea9ea6f1ebcR872, because there's no interpreter available to get a stack trace and it's not possible to return error from there. Sorry for that. 

Current version is good enough (the liftBitwise signature is a bit weird now, but it's not a big deal).

PR #422, opened 2020-06-12
perf(stdlib): ⚡️  implement manifestJsonEx as native Go

ghostsquad:
```
❯ make benchmark FILTER=Builtin_manifestJsonEx
go build ./cmd/jsonnet
./benchmark.sh Builtin_manifestJsonEx
Running Before Test... (10s)
Running After Test... (10s)
benchmark                               old ns/op     new ns/op     delta
Benchmark_Builtin_manifestJsonEx-16     22656394      7502016       -66.89%
```

---

EDIT
~I could use a small amount of help here. The test is currently failing seemingly because the output doesn't match. I suspect it's some sort of whitespace issue, but I can't tell.~

Fixed the problem, I was setting an initial indentation.

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -25,7 +25,9 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/google/go-jsonnet/ast"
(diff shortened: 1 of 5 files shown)


ghostsquad:
This should resolve this issue:
https://github.com/coreos/kube-prometheus/issues/537

on version 1.16.0 - takes 19s to render
after update - takes 2s to render

ghostsquad:
I fixed the whitespace problem.

builtins.go
```diff
@@ -1161,6 +1163,99 @@ func builtinParseJSON(i *interpreter, trace traceElement, str value) (value, err
 	return jsonToValue(i, trace, parsedJSON)
 }
 
+func builtinManifestJSONEx(i *interpreter, trace traceElement, obj, indent value) (value, error) {
```
sbarzowski, line 1166: We have a very similar logic here https://github.com/google/go-jsonnet/blob/v0.16.0/interpreter.go#L695 and here: https://github.com/google/go-jsonnet/blob/v0.16.0/interpreter.go#L627. 

Could you add a TODO to look into unifying the two? (Unless you feel like doing that now.) 

Please also add a comment that we implement it instead of using a library function to have full control over formatting.

builtins.go
```diff
@@ -1161,6 +1163,99 @@ func builtinParseJSON(i *interpreter, trace traceElement, str value) (value, err
...
+			return strings.Join(lines, ""), nil
+		case *valueObject:
+			newIndent := cindent + sindent
+			lines := []string{"{\n"}
+
+			fields := objectFields(v, withoutHidden)
+			sort.Strings(fields)
+			var objectLines []string
+			for _, fieldName := range fields {
+				fieldValue, err := v.index(i, trace, fieldName)
```
sbarzowski, line 1227: Please check the error.

ghostsquad: done

builtins.go
ghostsquad: I added a comment. I definitely think you are right, that these functions should be unified. I would have tackled this, but the use of `interface{}` in the above functions vs the `Value` based types might make it more difficult. I think that ValueArray and ValueObjects would need to be first converted to standard slice and map first, and iterated on again. I'm unsure if that's a significant performance impact or not.

builtins.go
```diff
@@ -1161,6 +1163,106 @@ func builtinParseJSON(i *interpreter, trace traceElement, str value) (value, err
...
+			lines := []string{"[\n"}
+
+			var arrayLines []string
+			for aI, cThunk := range v.elements {
+				cTv, err := cThunk.getValue(i, trace)
+				if err != nil {
+					return "", err
+				}
+
+				newPath := append(path, string(aI))
```
sbarzowski, line 1213: string(aI) doesn't do what you think. See: https://play.golang.org/p/UkFqprKKRm4. You probably want strconv.FormatInt(aI).

ghostsquad: fixed. Sounds like maybe I should write some tests!

ghostsquad:
@sbarzowski are you going to create a release for this?

sbarzowski:
Releases are not coupled with features, they are cut every couple of months. The last one was on May 22. Release process takes quite a lot of time (take a look at the release procedure: https://github.com/google/jsonnet/pull/802/files). I would like to automate some of that stuff eventually, but it's not easy (most steps may uncover problems requiring manual intervention). 

Anyway, depending on `master` version should be fairly safe, so for bleeding edge I recommend using that.

ghostsquad:
@sbarzowski holy cow. I had no idea that releases were that involved. Ok, I'll continue to recommend folks use the `go get -u` method of installing. If I'm not mistaken, that uses the latest on `master` not the releases?

sbarzowski:
> If I'm not mistaken, that uses the latest on master not the releases?

Yes, I think so.

PR #429, opened 2020-06-30
Reformat the --help text for readability

robx:
- separate the sections more cleanly, by adding empty lines and indenting
- align descriptions
- wrap long lines

The new help text is:

```
Jsonnet commandline interpreter v0.16.0

jsonnet {<option>} <filename>

Available options:
  -h / --help                This message
  -e / --exec                Treat filename as code
  -J / --jpath <dir>         Specify an additional library search dir
                             (right-most wins)
  -o / --output-file <file>  Write to the output file rather than stdout
  -m / --multi <dir>         Write multiple files to the directory, list files
                             on stdout
  -c / --create-output-dirs  Automatically creates all parent directories for
                             files
  -y / --yaml-stream         Write output as a YAML stream of JSON documents
  -S / --string              Expect a string, manifest as plain text
  -s / --max-stack <n>       Number of allowed stack frames
  -t / --max-trace <n>       Max length of stack trace before cropping
  --version                  Print version

Available options for specifying values of 'external' variables:
  Provide the value as a string:
  -V / --ext-str <var>[=<val>]      If <val> is omitted, get from environment
                                    var <var>
       --ext-str-file <var>=<file>  Read the string from the file
  Provide a value as Jsonnet code:
  --ext-code <var>[=<code>]         If <code> is omitted, get from environment
                                    var <var>
  --ext-code-file <var>=<file>      Read the code from the file

Available options for specifying values of 'top-level arguments':
  Provide the value as a string:
  -A / --tla-str <var>[=<val>]      If <val> is omitted, get from environment
                                    var <var>
       --tla-str-file <var>=<file>  Read the string from the file
  Provide a value as Jsonnet code:
  --tla-code <var>[=<code>]         If <code> is omitted, get from environment
                                    var <var>
  --tla-code-file <var>=<file>      Read the code from the file

Environment variables:
  JSONNET_PATH is a colon (semicolon on Windows) separated list of directories
  added in reverse order before the paths specified by --jpath (i.e. left-most
  wins). E.g. these are equivalent:
    JSONNET_PATH=a:b jsonnet -J c -J d
    JSONNET_PATH=d:c:a:b jsonnet
    jsonnet -J b -J a -J c -J d

In all cases:
  <filename> can be - (stdin)
  Multichar options are expanded e.g. -abc becomes -a -b -c.
  The -- option suppresses option processing for subsequent arguments.
  Note that since filenames and jsonnet programs can begin with -, it is
  advised to use -- if the argument is unknown, e.g. jsonnet -- "$FILENAME".
```


sbarzowski:
Cool, thanks. 

We'll need to change the expected output in the other repo:
* https://github.com/google/jsonnet/blob/master/test_cmd/no_args.golden.stderr.golang
* https://github.com/google/jsonnet/blob/master/test_cmd/help.golden.stdout.golang

Can you update these files? Could you also check if the help formatting in `jsonnetfmt` command is consistent with your changes (https://github.com/google/go-jsonnet/blob/master/cmd/jsonnetfmt/cmd.go)?

cmd/jsonnetfmt/cmd.go
```diff
@@ -38,28 +38,34 @@ func usage(o io.Writer) {
...
+	fmt.Fprintln(o, "  -h / --help                This message")
+	fmt.Fprintln(o, "  -e / --exec                Treat filename as code")
+	fmt.Fprintln(o, "  -o / --output-file <file>  Write to the output file rather than stdout")
+	fmt.Fprintln(o, "  -c / --create-output-dirs  Automatically creates all parent directories for")
+	fmt.Fprintln(o, "                             files")
+	fmt.Fprintln(o, "  -i / --in-place            Update the Jsonnet file(s) in place")
+	fmt.Fprintln(o, "  --test                     Exit with failure if reformatting changed the")
+	fmt.Fprintln(o, "                             file(s)")
+	fmt.Fprintln(o, "  -n / --indent <n>          Number of spaces to indent by")
+	fmt.Fprintln(o, "                             (default 2, 0 means no change)")
```
robx, line 50: not really sure whether this is better than breaking in the middle of the "default" thingy

sbarzowski: It's ok.

sbarzowski:
Thanks!

PR #442, opened 2020-07-28
issue-433 increase the number of constructed Jsonnet handles.

### LINKED ISSUE #433: maximum number of constructed Jsonnet handles exceeded
Hi,

I have a very large project using cpp-jsonnet with Kapitan. When attempting using go-jsonnet via python c-bindings (instead of cpp-jsonnet), I keep getting max handles exceed error message:
```maximum number of constructed Jsonnet handles exceeded ```

However, having manually updated this value (to a finger in the air value `const maxID = 1000000`) and recompiled go-jsonnet, the error goes away and everything works.

Not sure if the fault is at my side, but is there any chance the default number of handles can be increased? Or perhaps offer a way to specify a higher value?

### PR DESCRIPTION
**Pull request by alldroll**
It resolves #433

### DIFF
--- a/c-bindings/handles.go
+++ b/c-bindings/handles.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 )
 
-const maxID = 100000
+const maxID = (1 << 63) - 1
 
 // Because of Go GC, there are restrictions on keeping Go pointers in C.
 // We cannot just pass *jsonnet.VM/JsonValue to C. So instead we use "handle" structs in C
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by sbarzowski** [comment]
Looks good, but I want to read it more carefully. 

Eventually I would like to avoid keeping all the allocated handles forever, but this is already a big improvement. Avoiding the "leak" will probably require using the map or something, instead of the flat list, which may degrade performance a little bit, but it shouldn't matter too much.

File: c-bindings/handles.go [hunk 1]
```diff
@@ -17,62 +15,45 @@ const maxID = 100000
...
-
-	return h.objects[id-1], nil
-}
-
-// ensureValidID returns an error if the given handle ID is invalid, otherwise returns nil
-func (h *handlesTable) ensureValidID(id uint32) error {
-	if id == 0 || uint64(id) > uint64(len(h.objects)) {
-		return errInvalidHandle
+func (h *handlesTable) get(id uintptr) (interface{}, error) {
+	if _, ok := h.handles[id]; !ok {
```
sbarzowski, line 54: Nit: This function indexes the map twice – once to check and then again to get the value. It would be enough to have `handle, ok := h.handles[id]` before this if.

Issue #329, opened 2019-10-06
Establish sentinel value for Importer.Import's "importedFrom" parameter


seh:
Per preceding discussion in bitnami/kubecfg#266, the `Importer.Importer` method's "importedFrom" parameter is not documented at present, and interpreting its value varies by `Importer` implementation. When importing values for external variables and TLAs, though, such as for files nominated either by the `--ext-str-file`, `--ext-code-file`, `--tla-str-file`, and `--tla-code-file` flags or in the value of the `--ext-code` or `--tla-code` flags, the "importedFrom" location is best thought of as "the top-level context of program," which often means the current working directory.

In order to allow `Importer` implementations to detect these special importing scenarios, @sbarzowski proposed replacing the current sentinel values passed as arguments for the "importedFrom" parameter (e.g. `<extvar:example>`) with **the empty string**, per https://github.com/bitnami/kubecfg/pull/266#issuecomment-535090890. Not just that, but we would also document this contract as part of the `Importer` interface, so that all `Importer` implementations could count on receiving this value to trigger their own interpretation of "this path is being imported in the top-level context."

The pending bitnami/kubecfg#271 shows an `Importer` implementation that depends upon this proposed contract.

seh:
In the current implementation, function `prepareExtVars` in file _interpreter.go_ [prepares the current pseudo-file names like `<extvar:example>`](https://github.com/google/go-jsonnet/blob/56da397/interpreter.go#L1131) and passes them down through the the code-to-AST gullet. It's useful to have these indicative placeholders available in some places for diagnostic clarity, but we need to be able to tell later that they're not actual file paths. Hence, we could consider expanding `program.SnippetToAST` to take an additional parameter that distinguishes between the source "file name" (again, not always nominating a real file) and the value suitable for passing to `Importer.Import`'s "importedFrom" parameter.

Changing the signature of the exported function `program.SnippetToAST` is not nice for existing callers, but `program` is an internal package, so I think we _do_ have that freedom.

seh:
Though not directly related to the proposal here, @sbarzowski also noted that these top-level imports from these pseudo-files should not be cached (https://github.com/bitnami/kubecfg/pull/266#issuecomment-534892902).

sbarzowski:
> Though not directly related to the proposal here, @sbarzowski also noted that these top-level imports from these pseudo-files should not be cached (bitnami/kubecfg#266 (comment)).

[...]

PR #443, opened 2020-08-03
jsonnet-deps: Jsonnet static dependency parser

varunbpatil:
Fixes https://github.com/google/jsonnet/issues/833.

### DIFF
--- a/cmd/jsonnet-deps/cmd.go
+++ b/cmd/jsonnet-deps/cmd.go
@@ -1,0 +1,208 @@
+/*
+Copyright 2020 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/fatih/color"
+
+	"github.com/google/go-jsonnet"
+	"github.com/google/go-jsonnet/cmd/internal/cmd"
+)
+
+func version(o io.Writer) {
+	fmt.Fprintf(o, "Jsonnet static dependency parser %s\n", jsonnet.Version())
+}
+
+func usage(o io.Writer) {
+	version(o)
+	fmt.Fprintln(o)
+	fmt.Fprintln(o, "jsonnet-deps {<option>} <filename>...")
+	fmt.Fprintln(o)
+	fmt.Fprintln(o, "Available options:")
+	fmt.Fprintln(o, "  -h / --help                This message")
+	fmt.Fprintln(o, "  -J / --jpath <dir>         Specify an additional library search dir")
+	fmt.Fprintln(o, "                             (right-most wins)")
+	fmt.Fprintln(o, "  -o / --output-file <file>  Write to the output file rather than stdout")
+	fmt.Fprintln(o, "  --version                  Print version")
+	fmt.Fprintln(o)
+	fmt.Fprintln(o, "Environment variables:")
+	fmt.Fprintln(o, "  JSONNET_PATH is a colon (semicolon on Windows) separated list of directories")
+	fmt.Fprintln(o, "  added in reverse order before the paths specified by --jpath (i.e. left-most")
+	fmt.Fprintln(o, "  wins). E.g. these are equivalent:")
+	fmt.Fprintln(o, "    JSONNET_PATH=a:b jsonnet -J c -J d")
+	fmt.Fprintln(o, "    JSONNET_PATH=d:c:a:b jsonnet")
+	fmt.Fprintln(o, "    jsonnet -J b -J a -J c -J d")
+	fmt.Fprintln(o)
+	fmt.Fprintln(o, "In all cases:")
+	fmt.Fprintln(o, "  Multichar options are expanded e.g. -abc becomes -a -b -c.")
+	fmt.Fprintln(o, "  The -- option suppresses option processing for subsequent arguments.")
+	fmt.Fprintln(o, "  Note that since filenames and jsonnet programs can begin with -, it is")
+	fmt.Fprintln(o, "  advised to use -- if the argument is unknown, e.g. jsonnet-deps -- \"$FILENAME\".")
+}
+
+type config struct {
+	inputFiles []string
+	outputFile string
+	jPaths     []string
+}
+
+type processArgsStatus int
+
+const (
+	processArgsStatusContinue     = iota
+	processArgsStatusSuccessUsage = iota
+	processArgsStatusFailureUsage = iota
+	processArgsStatusSuccess      = iota
+	processArgsStatusFailure      = iota
+)
+
+func processArgs(givenArgs []string, conf *config, vm *jsonnet.VM) (processArgsStatus, error) {
+	args := cmd.SimplifyArgs(givenArgs)
+	remainingArgs := make([]string, 0, len(args))
+
+	for i := 0; i < len(args); i++ {
+		arg := args[i]
+		if arg == "-h" || arg == "--help" {
+			return processArgsStatusSuccessUsage, nil
+		} else if arg == "-v" || arg == "--version" {
+			version(os.Stdout)
+			return processArgsStatusSuccess, nil
+		} else if arg == "-o" || arg == "--output-file" {
+			outputFile := cmd.NextArg(&i, args)
+			if len(outputFile) == 0 {
+				return processArgsStatusFailure, fmt.Errorf("-o argument was empty string")
+			}
+			conf.outputFile = outputFile
+		} else if arg == "-J" || arg == "--jpath" {
+			dir := cmd.NextArg(&i, args)
+			if len(dir) == 0 {
+				return processArgsStatusFailure, fmt.Errorf("-J argument was empty string")
+			}
+			conf.jPaths = append(conf.jPaths, dir)
+		} else if arg == "--" {
+			// All subsequent args are not options.
+			i++
+			for ; i < len(args); i++ {
+				remainingArgs = append(remainingArgs, args[i])
+			}
+			break
+		} else if len(arg) > 1 && arg[0] == '-' {
+			return processArgsStatusFailure, fmt.Errorf("unrecognized argument: %s", arg)
+		} else {
+			remainingArgs = append(remainingArgs, arg)
+		}
+	}
+
+	if len(remainingArgs) == 0 {
+		return processArgsStatusFailureUsage, fmt.Errorf("must give filename")
+	}
+	conf.inputFiles = remainingArgs
+
+	return processArgsStatusContinue, nil
+}
+
+func writeDependencies(dependencies []string, outputFile string) (err error) {
+	var f *os.File
+
+	if outputFile == "" {
+		f = os.Stdout
+	} else {
+		f, err = os.Create(outputFile)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if ferr := f.Close(); ferr != nil {
+				err = ferr
+			}
+		}()
+	}
+
+	if len(dependencies) != 0 {
+		output := strings.Join(dependencies, "\n") + "\n"
+		_, err = f.WriteString(output)
+		if err != nil {
+			return err
+		}
+	}
+
+	return
+}
+
+func main() {
+	cmd.StartCPUProfile()
+	defer cmd.StopCPUProfile()
+
+	vm := jsonnet.MakeVM()
+	vm.ErrorFormatter.SetColorFormatter(color.New(color.FgRed).Fprintf)
+
+	conf := config{}
+	jsonnetPath := filepath.SplitList(os.Getenv("JSONNET_PATH"))
+	for i := len(jsonnetPath) - 1; i >= 0; i-- {
+		conf.jPaths = append(conf.jPaths, jsonnetPath[i])
+	}
+
+	status, err := processArgs(os.Args[1:], &conf, vm)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "ERROR: "+err.Error())
+	}
+	switch status {
+	case processArgsStatusContinue:
+		break
+	case processArgsStatusSuccessUsage:
+		usage(os.Stdout)
+		os.Exit(0)
+	case processArgsStatusFailureUsage:
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "")
+		}
+		usage(os.Stderr)
+		os.Exit(1)
+	case processArgsStatusSuccess:
+		os.Exit(0)
+	case processArgsStatusFailure:
+		os.Exit(1)
+	}
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: conf.jPaths})
+
+	for _, file := range conf.inputFiles {
+		if _, err := os.Stat(file); err != nil {
+			fmt.Fprintln(os.Stderr, err.Error())
+			os.Exit(1)
+		}
+	}
+
+	dependencies, err := vm.FindDependencies("", conf.inputFiles)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+	cmd.MemProfile()
+
+	err = writeDependencies(dependencies, conf.outputFile)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err.Error())
+		os.Exit(1)
+	}
+}
--- a/cmd/jsonnet/cmd.go
+++ b/cmd/jsonnet/cmd.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/fatih/color"
@@ -91,15 +90,6 @@ func usage(o io.Writer) {
 	fmt.Fprintln(o, "  The -- option suppresses option processing for subsequent arguments.")
 	fmt.Fprintln(o, "  Note that since filenames and jsonnet programs can begin with -, it is")
 	fmt.Fprintln(o, "  advised to use -- if the argument is unknown, e.g. jsonnet -- \"$FILENAME\".")
-}
-
-func safeStrToInt(str string) (i int) {
-	i, err := strconv.Atoi(str)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Invalid integer \"%s\"\n", str)
-		os.Exit(1)
-	}
-	return
 }
 
 type config struct {
@@ -203,7 +193,7 @@ func processArgs(givenArgs []string, config *config, vm *jsonnet.VM) (processArg
 			}
 			break
 		} else if arg == "-s" || arg == "--max-stack" {
-			l := safeStrToInt(cmd.NextArg(&i, args))
+			l := cmd.SafeStrToInt(cmd.NextArg(&i, args))
 			if l < 1 {
 				return processArgsStatusFailure, fmt.Errorf("invalid --max-stack value: %d", l)
 			}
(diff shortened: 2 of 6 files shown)


varunbpatil:
@googlebot I signed it!

sbarzowski:
Thanks for the contribution! 

Don't worry about the failing regression test. We will need to refactor it a bit anyway. Most importantly, it should actually be a separate command and not just an option to the interpreter command, because it does not run any Jsonnet code and most of the other options are not relevant for it.

Please let me know if you want to continue working on that (I will do a detailed review then) or if you want me to take over from here. Either way works for me.

varunbpatil:
Thanks, @sbarzowski. The new command makes sense. I will work on that.
Would a command named jsonnetdeps be okay with the following arguments carried over from jsonnet cli?
```
-h / --help
-e / --exec
-J / --jpath
-o / --output-file
-s / --max-stack
-t / --max-trace
```

varunbpatil:
Hi @sbarzowski, I have created a new command as suggested. I'm not very familiar with the Bazel build system, so I might have missed a few things w.r.t to the Bazel build. Kindly share your inputs.

cmd/jsonnet/cmd.go
```diff
@@ -314,7 +314,7 @@ func writeMultiOutputFiles(output map[string]string, outputDir, outputFile strin
 			return err
 		}
 		defer func() {
-			if ferr := manifest.Close(); err != nil {
+			if ferr := manifest.Close(); ferr != nil {
```
sbarzowski, line 317: Good catch!

cmd/jsonnetdeps/cmd.go
```diff
@@ -0,0 +1,252 @@
...
+		} else if arg == "-J" || arg == "--jpath" {
+			dir := cmd.NextArg(&i, args)
+			if len(dir) == 0 {
+				return processArgsStatusFailure, fmt.Errorf("-J argument was empty string")
+			}
+			if dir[len(dir)-1] != '/' {
+				dir += "/"
+			}
+			conf.jPaths = append(conf.jPaths, dir)
+		} else if arg == "-s" || arg == "--max-stack" {
```
sbarzowski, line 122: This can be dropped – it only makes sense for the actual evaluation. It limits the depth of non-tailstrict recursion to prevent Jsonnet from eating all your memory if there is a bug.

cmd/jsonnetdeps/cmd.go
```diff
@@ -0,0 +1,252 @@
...
+			l := safeStrToInt(cmd.NextArg(&i, args))
+			if l < 1 {
+				return processArgsStatusFailure, fmt.Errorf("invalid --max-stack value: %d", l)
+			}
+			vm.MaxStack = l
+		} else if arg == "-t" || arg == "--max-trace" {
```
sbarzowski, line 128: This also can be dropped, because it's relevant only to the runtime. This specifies the maximum number of printed stack frames if a Jsonnet program crashes (so that if it fails inside some deep recursion it won't flood the output with thousands of stack frames).

cmd/jsonnetdeps/cmd.go
```diff
@@ -0,0 +1,252 @@
...
+		os.Exit(1)
+	case processArgsStatusSuccess:
+		os.Exit(0)
+	case processArgsStatusFailure:
+		os.Exit(1)
+	}
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: conf.jPaths})
+
+	if len(conf.inputFiles) != 1 {
```
sbarzowski, line 230: Can you change the representation so that conf can only have one file (i.e. `inputFile []string` -> `inputFile string`)? It's good to make invalid states unrepresentable.

Alternatively you can actually support passing multiple files. It would then produce union of depeendency sets of the passed files.

vm.go
```diff
@@ -186,6 +187,41 @@ func (vm *VM) evaluateSnippet(filename string, snippet string, kind evalKind) (o
 	return output, nil
 }
 
+func (vm *VM) evaluateDependencies(fileName string, children []ast.Node, dependencies *[]string, stackTrace *[]traceFrame, recursionDepth int) (err error) {
```
sbarzowski, line 190: We should change the name here. Evaluate has a very specific meaning in Jsonnet. It means to take an expression (or a thunk or something) and produce a Jsonnet value according to the rules in https://jsonnet.org/ref/spec.html. 

Maybe `findDependencies`?

vm.go
```diff
@@ -186,6 +187,41 @@ func (vm *VM) evaluateSnippet(filename string, snippet string, kind evalKind) (o
...
+	if recursionDepth > vm.MaxStack {
```
sbarzowski, line 191: I think we can drop this check. This is only helpful to prevent infinite loop in case of recursive imports, but there is a better way. We can just check if we visited an import already and skip it (it's enough to process each import only once). What we're doing here is basically DFS.

This will simplify things a bit – we no longer need to care about stackTrace at all.

vm.go
```diff
@@ -225,6 +261,23 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
...
+// EvaluateSnippetDependencies evaluates a string containing Jsonnet code returning an array of dependencies.
+//
+// The filename parameter is only used for error messages.
+func (vm *VM) EvaluateSnippetDependencies(filename string, snippet string) (dependencies []string, formattedErr error) {
+	node, err := SnippetToAST(filename, snippet)
+	if err != nil {
+		return nil, err
+	}
+	var stackTrace []traceFrame
+	err = vm.evaluateDependencies(filename, parser.Children(node), &dependencies, &stackTrace, 0)
```
sbarzowski, line 273: What do you think about sorting the dependencies before returning them (with `sort.Strings`)?

This way we could make the output more stable (removing one dependency wouldn't change the relative order of anything that remains).

vm.go
```diff
@@ -225,6 +261,23 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
 	return
 }
 
+// EvaluateSnippetDependencies evaluates a string containing Jsonnet code returning an array of dependencies.
+//
+// The filename parameter is only used for error messages.
+func (vm *VM) EvaluateSnippetDependencies(filename string, snippet string) (dependencies []string, formattedErr error) {
```
sbarzowski, line 267: Same thing about this name – it doesn't really evaluate Jsonnet code. Maybe just FindDependencies?

vm.go
```diff
@@ -225,6 +261,23 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
...
+	node, err := SnippetToAST(filename, snippet)
+	if err != nil {
+		return nil, err
+	}
+	var stackTrace []traceFrame
+	err = vm.evaluateDependencies(filename, parser.Children(node), &dependencies, &stackTrace, 0)
+	if err != nil {
+		err = makeRuntimeError(err.Error(), stackTrace)
```
sbarzowski, line 275: Not sure about error formatting here, especially about wrapping an error from vm.ResolveImport in a RuntimeError. How does the output look like if you run the command on a file with missing import (e.g. `import "foo.libsonnet"` when there is not `foo.libsonnet`)?

cmd/jsonnetdeps/cmd.go
```diff
@@ -0,0 +1,252 @@
...
+)
+
+func version(o io.Writer) {
+	fmt.Fprintf(o, "Jsonnet static dependency parser %s\n", jsonnet.Version())
+}
+
+func usage(o io.Writer) {
+	version(o)
+	fmt.Fprintln(o)
+	fmt.Fprintln(o, "jsonnetdeps {<option>} <filename>")
```
sbarzowski, line 40: We don't have an established convention for naming the tools yet. What do you think about adding a dash, like `jsonnet-deps`, `jsonnet-lint`? It seems more readable to me than `jsonnetdeps`, `jsonnetlint`. The commands in the second style kinda blend together for me, especially with variable width font: jsonnetdeps, jsonnetlint.

I'm not sure. What do you think?

(Either way we're going to leave `jsonnetfmt` alone for now).

vm.go
```diff
@@ -225,6 +261,23 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
 	return
 }
 
+// EvaluateSnippetDependencies evaluates a string containing Jsonnet code returning an array of dependencies.
```
sbarzowski, line 264: It does not evaluate Jsonnet code. It finds all transitive dependencies (imports) of a Jsonnet program.

cmd/jsonnet/cmd.go
varunbpatil: Thanks!

cmd/jsonnetdeps/cmd.go
varunbpatil: Done.

cmd/jsonnetdeps/cmd.go
varunbpatil: Done.

vm.go
varunbpatil: Changed to `findDependencies`.

vm.go
varunbpatil: Changed to `FindDependencies`.

vm.go
varunbpatil: Done.

vm.go
varunbpatil: Makes sense. Done.

vm.go
varunbpatil:
```
$ cat bar.jsonnet
local foo = import 'foo.jsonnet';
foo

$ ~/go/bin/jsonnetdeps bar.jsonnet
RUNTIME ERROR: couldn't open import "foo.jsonnet": no match locally or in the Jsonnet library paths
        bar.jsonnet:1:13-33
```

varunbpatil: I was not sure whether missing imports should raise an error. One might want to simply ignore missing imports if they are wrapped in conditionals or such.

cmd/jsonnetdeps/cmd.go
varunbpatil: Absolutely. `jsonnet-deps` was my first choice as well. I changed my mind after looking at `jsonnetfmt`.

vm.go
varunbpatil: Ah. I hadn't thought about infinite loops. I'll fix it.

cmd/jsonnetdeps/cmd.go
varunbpatil: Updated to handle multiple input files.

vm.go
varunbpatil: Fixed.

vm.go
```diff
@@ -186,6 +188,42 @@ func (vm *VM) evaluateSnippet(filename string, snippet string, kind evalKind) (o
 	return output, nil
 }
 
+func (vm *VM) findDependencies(fileName string, children []ast.Node, dependencies *map[string]struct{}, stackTrace *[]traceFrame) (err error) {
```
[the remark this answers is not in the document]
varunbpatil: Thanks. Fixed.

cmd/jsonnet-deps/cmd.go
```diff
@@ -0,0 +1,225 @@
...
+		os.Exit(1)
+	}
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: conf.jPaths})
+
+	deps := make(map[string]struct{})
+	for _, filename := range conf.inputFiles {
+		input := cmd.SafeReadInput(conf.filenameIsCode, &filename)
+		dependencies, err := vm.FindDependencies(filename, input)
+		cmd.MemProfile()
```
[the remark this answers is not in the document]
varunbpatil: Done. Moved it to a point just before writing to the output file.

vm.go
```diff
@@ -225,6 +263,31 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
...
+//
+// The filename parameter is only used for error messages.
+func (vm *VM) FindDependencies(filename string, snippet string) ([]string, error) {
+	node, err := SnippetToAST(filename, snippet)
+	if err != nil {
+		return nil, err
+	}
+	deps := make(map[string]struct{})
+	var stackTrace []traceFrame
+	err = vm.findDependencies(filename, parser.Children(node), &deps, &stackTrace)
```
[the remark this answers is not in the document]
varunbpatil: Thanks. I felt I was somehow skipping a level when I directly jumped to the children, but couldn't put my finger on it. Fixed.

cmd/jsonnet-deps/cmd.go
```diff
@@ -0,0 +1,225 @@
...
+	case processArgsStatusSuccess:
+		os.Exit(0)
+	case processArgsStatusFailure:
+		os.Exit(1)
+	}
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: conf.jPaths})
+
+	deps := make(map[string]struct{})
+	for _, filename := range conf.inputFiles {
```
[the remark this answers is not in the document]
varunbpatil: Done.

cmd/jsonnet-deps/cmd.go
```diff
@@ -0,0 +1,225 @@
...
+		os.Exit(1)
+	case processArgsStatusSuccess:
+		os.Exit(0)
+	case processArgsStatusFailure:
+		os.Exit(1)
+	}
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: conf.jPaths})
+
+	deps := make(map[string]struct{})
```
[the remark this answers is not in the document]
varunbpatil: Done.

vm.go
```diff
@@ -225,6 +263,31 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
...
+// FindDependencies finds all transitive dependencies (import and importstr) of a snippet of
+// Jsonnet code and returns a sorted array of unique dependencies.
+//
+// The filename parameter is only used for error messages.
+func (vm *VM) FindDependencies(filename string, snippet string) ([]string, error) {
+	node, err := SnippetToAST(filename, snippet)
+	if err != nil {
+		return nil, err
+	}
+	deps := make(map[string]struct{})
```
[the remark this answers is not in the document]
varunbpatil: Done, although I have some questions on the `foundAt` filename which I've put in the other open comment.

vm.go
```diff
@@ -225,6 +263,31 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
 	return
 }
 
+// FindDependencies finds all transitive dependencies (import and importstr) of a snippet of
+// Jsonnet code and returns a sorted array of unique dependencies.
+//
+// The filename parameter is only used for error messages.
+func (vm *VM) FindDependencies(filename string, snippet string) ([]string, error) {
```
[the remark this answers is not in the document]
varunbpatil: Makes sense. I will remove support for `--exec` and `stdin`, and instead pass all the input files to `FindDependencies` where I can loop over them while keeping a common dependencies map.

I have a few questions:

> taking a filePath (or multiple filePaths) and importedFrom here

What would the value of `importedFrom` be for the input files? If it is always some dummy file in the current working directory, can it be dropped from the arguments to `FindDependencies`?

> This would give us the canonical path

The current implementation of `Import` (for FIleImporter) seems to return relative paths when either `importedFrom` or `--jpath` is a relative path. However, looking at `importer.tryPath()` which returns absPath, the expectation is that it should return an absolute path. So, should I fix `importer.tryPath()` to actually return the absolute path?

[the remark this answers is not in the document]
varunbpatil: The current `Import` implementation also treats `/a/b/test.libsonnet` different from `/a/b/c/../test.libsonnet`, but they are actually the same file. `filepath.Abs()` handles this case, but doesn't work with symlinks.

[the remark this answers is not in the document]
sbarzowski: > What would the value of importedFrom be for the input files? If it is always some dummy file in the current working directory, can it be dropped from the arguments to FindDependencies? 

It should probably be `""` (an empty string). I believe it will be interpreted as CWD. Eventually, we want to generalize "" as "top-level-import", per #329.

I would keep the `importedFrom` in FindDependencies and pass `""` from cmd. This is consistent with other functions and provides some extra flexibility.

> The current implementation of `Import` (for FIleImporter) seems to return relative paths when either `importedFrom` or `--jpath` is a relative path. However, looking at `importer.tryPath()` which returns absPath, the expectation is that it should return an absolute path. So, should I fix `importer.tryPath()` to actually return the absolute path?

You're right and I didn't notice that before. It should always return the absolute path.

varunbpatil:
I don't know what the expected behavior of `std.thisFile` is. According to https://github.com/google/jsonnet/issues/699, it should be a relative path, but, my changes to `Import` to return absolute path will break this. Also, I seem to be getting two different behaviors w.r.t `std.thisFile`. When importing, I get the absolute path (because of my changes), but when accessing as a field, I get a relative path.

Not sure where to go from here. I also don't understand the "hermeticity" claim in the context of `std.thisFile` if it is expected to  return a relative path because the path from where jsonnet is executed should also be considered as "execution environment" and therefore should be independent of it.

sbarzowski:
Thanks for being thorough here!

> Not sure where to go from here. I also don't understand the "hermeticity" claim in the context of `std.thisFile` if it is expected to return a relative path because the path from where jsonnet is executed should also be considered as "execution environment" and therefore should be independent of it.

Yes, `std.thisFile` is in a gray area and I would argue against including it, if we were designing the language again. It leaks something which should be an implementation detail of the importer. Right now, when thinking rigorously about what Jsonnet evaluations depends on, we need to include paths to the files.

This isn't *completely* terrible, because (a) it's still fairly limited information, which we can precisely describe (b) it's still possible and practical to have full control over there vales when using a custom importer (c) people don't seem to abuse it too badly in practice. 

> According to google/jsonnet#699, it should be a relative path, but, my changes to Import to return absolute path will break this. 

I think what @sparkprime meant was that **if** it's relative, then it's relative to the original cwd. Even now, it doesn't have to be relative (for example when all the paths were passed as absolute). So, I think it's fine to have absolute paths there.

> Also, I seem to be getting two different behaviors w.r.t `std.thisFile`. When importing, I get the absolute path (because of my changes), but when accessing as a field, I get a relative path.
> 

This is actually a deeper issue. The filepath that is passed to CLI is an ordinary filepath. We read the file outside of the importer and then pass it to EvaluateSnippet and hope that the path will be compatible with what the importer expects as a canonical path. Also this initial file is not cached, which is also not ideal.

I think we should unify it and always use the importer for reading paths. We already do that for paths in `--tla-code-file` etc, so why not for the initial file?

I want to fix this, but it's clearly out of scope for this PR :-). Don't worry about it here.

sbarzowski:
I looked at the tests now, and I see that absolute files can get a bit ugly with Bazel and you needed to hack the existing ones with regexps. And now it won't work with `go test -update`. Hmmmm...

It's getting a bit out of hand here. Perhaps it would be better to revert the absolute path change for now. I will think more about this and we'll figure something out. I don't want this controversial part to block the rest of this PR. Sorry for the trouble.

vm.go
```diff
@@ -225,6 +263,53 @@ func (vm *VM) EvaluateSnippetMulti(filename string, snippet string) (files map[s
...
+	deps := make(map[string]struct{})
+
+	for i, filePath := range importedPaths {
+		node, foundAt, err := vm.ImportAST(importedFrom, filePath)
+		if err != nil {
+			return nil, err
+		}
+
+		// Overwrite `importedPaths` with their canonical paths.
+		importedPaths[i] = foundAt
```
sbarzowski, line 281: The slices are mutable – this changes the argument that was passed to us. It can be very confusing for the user of this library. Create a new slice instead.

varunbpatil: Got it.

varunbpatil:
> It's getting a bit out of hand here. Perhaps it would be better to revert the absolute path change for now. I will think more about this and we'll figure something out. I don't want this controversial part to block the rest of this PR. Sorry for the trouble.

No problem. I agree. I thought of keeping just the symlink evaluation in `imports.go` so that symlinks are handled correctly, but it seems Bazel test data relies on symbolic links and so only the "Bazel Check" fails while `go test ./...` passes. So, I've reverted `imports.go` completely.

In `(f|F)indDependencies`, I'm calculating the absolute path from `foundAt` and putting the absolute path into the dependencies map to satisfy the uniqueness constraint. However, I'm still passing `foundAt` to `importedFrom` in the recursion. Please let me know if this is okay.

sbarzowski:
Looks good, I'm going to merge this now. Thanks a lot for contributing a whole new tool. I appreciate the willingness to dive deep – it's very important when working on programming language tools.

> In (f|F)indDependencies, I'm calculating the absolute path from foundAt (if using FileImporter) and putting the absolute path into the dependencies map to satisfy the uniqueness constraint. However, I'm still passing foundAt to importedFrom in the recursion. Please let me know if this is okay.

That's a reasonable choice for now.

varunbpatil:
Thanks for the guidance throughout the process @sbarzowski. My first contribution in Golang :-) I'm glad to have collaborated with you.

PR #444, opened 2020-08-08
Add github action to build binaries with goreleaser.

PaulRudin:
This add a github action to build binaries and add them as release artefacts using goreleaser (https://goreleaser.com/).

It seems to work fine - see e.g. https://github.com/PaulRudin/go-jsonnet/releases/tag/v0.0.200. I've checked that the binaries for linux amd64 work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+# build binaries and add to release - see
+# https://github.com/goreleaser/goreleaser-action
+
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


PaulRudin:
Dunno why it's doing the cla: no thing - I've signed before - I'll sort it out.

PaulRudin:
Ah - wrong email address on the commit. Fixed now.

sbarzowski:
Thanks a lot. Let's give this a try. It's nice that it is simpler than what we did with TravisCI in https://github.com/google/jsonnet.

PaulRudin:
Yeah, there were no github actions then. 

You don't need to faff about with passing credentials when it's all within github, and there are loads of off the shelf actions for common stuff.

PR #447, opened 2020-08-16
Anonymous Import Locations

### LINKED ISSUE #329: Establish sentinel value for Importer.Import's "importedFrom" parameter

sbarzowski:
We used to treat dummy paths like `<stdin>`, `<std>`, `<extvar>` as
real import locations, which causes obvious problem for importers.

After this change we consistently pass "" (an empty string) as location
for non-imported paths.

We exposed new functions to properly handle all paths.

The first function family, EvaluateFile*  allows evaluating
a Jsonnet program at a specified importable path. It should
be used in most cases.

The second function family, EvaluateAnonymousSnippet* allows evaluating
a snippet without an importable path. It should be used for situations
like the code passed as a commandline argument, stdin, etc.

The old function family, EvaluateSnippet* is now deprecated, but
works the same as before, i.e. the passed filenames are treated
as imported paths.

Changes are required to custom importers to make sure they satisfy
the refined contract.

Fixes #329.

### DIFF
--- a/ast/location.go
+++ b/ast/location.go
@@ -21,9 +21,17 @@ import (
 	"fmt"
 )
 
+// DiagnosticFileName is a file name used for diagnostics.
+// It might be a dummy value, such as <std> or <extvar:something>.
+// It should never be passed to an importer.
+type DiagnosticFileName string
+
 // Source represents a source file.
 type Source struct {
 	Lines []string
+	// DiagnosticFileName is the imported path or a special string
+	// for indicating stdin, extvars and other non-imported sources.
+	DiagnosticFileName DiagnosticFileName
 }
 
 //////////////////////////////////////////////////////////////////////////////
@@ -60,6 +68,7 @@ func LocationBefore(a Location, b Location) bool {
 
 // LocationRange represents a range of a source file.
 type LocationRange struct {
+	// FileName should be the imported path or "" for snippets etc.
 	FileName string
 	Begin    Location
 	End      Location // TODO(sbarzowski) inclusive? exclusive? a gap?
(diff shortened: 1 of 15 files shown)


sbarzowski:
@seh @sh0rez

sparkprime:
So I guess most importers probably just took dirname of the path, which is "" for both inputs "" and "<something>" so this is a pretty safe change?

sbarzowski:
> So I guess most importers probably just took dirname of the path, which is "" for both inputs "" and "<something>" so this is a pretty safe change?

Yes, this is true, for our FileImporter at least. Dealing with "<something>" caused trouble for 3rd party importers.

sparkprime:
LGTM

PR #459, opened 2020-09-29
Add debian package build to goreleaser

sargun:
This makes the goreleaser output two debian packages. One for jsonnet,
and another for jsonnetfmt. Since we don't use cgo, we don't need to
take dependencies, as we're fully statically linked. The package
name for this package will be go-jsonnet, and go-jsonnetfmt. There
is a package that Ubuntu has released that's based on the C++
version. We have a conflict statement with it in order to
avoid people installing both.

    We can ask the C++ package to rename itself to jsonnet-c++, and then
    add a provides for the package name jsonnet. The reason to not name
    this jsonnet immediately, even though the Go version if meant to
    supercede the C++ version, is that this doesn't fully implement the
    C++ version (for example, man pages).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -51,3 +51,37 @@ archives:
       amd64: x86_64
 checksum:
   name_template: 'checksums.txt'
+
+nfpms:
+  - id: jsonnet
+    package_name: jsonnet-go
+    builds:
+      - jsonnet
+    description: A data templating language for app and tool developers
+    homepage: https://github.com/google/go-jsonnet
+    license: Apache 2.0
+    formats:
+      - deb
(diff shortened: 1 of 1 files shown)


sargun:
@googlebot I signed it!

sargun:
I'm not sure why the travis build failed with:

```
ERRO Running error: context loading failed: no go files to analyze 

````

sbarzowski:
Thanks, looks good. The CI issue doesn't seem related to this change. I'll look into it later.

sargun:
Anything you need / want done prior to moving forward?

sbarzowski:
Thanks for asking. I just need to sort out the CI issue. No further action is needed on your end.

PR #466, opened 2020-10-25
Add support for missing functions in C-bindings.

sbarzowski:
With this change we cover 100% of the C API.

### DIFF
--- a/c-bindings-tests/compat_test.py
+++ b/c-bindings-tests/compat_test.py
@@ -17,6 +17,36 @@ lib.jsonnet_evaluate_snippet.argtypes = [
 ]
 lib.jsonnet_evaluate_snippet.restype = ctypes.POINTER(ctypes.c_char)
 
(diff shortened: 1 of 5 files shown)


sbarzowski:
@alldroll @ramaro @sparkprime

sbarzowski:
(After this change they should also work with Python bindings).

alldroll:
LGTM!

Could you please describe why is [it](https://github.com/google/go-jsonnet/pull/466/files#diff-d239e67b7d1d363de0b39f18b145f425b4ddc093e59304d55cc5c86d592e6c47R249) a concern for you?

sbarzowski:
Thanks. The concern was about how it's supposed to be freed. I verified that it's supposed to be freed using jsonnet_realloc (that's what the comments in libjsonnet.h require) which results in it being freed with `free` and that's what memory allocated with `C.Cstring` expects, so we're good.

PR #470, opened 2020-11-15
Better handling of stack traces.

sbarzowski:
Fixes multiple issues with stack traces leading to missing
stack trace lines. Also, we no longer put builtin context
on the line which *calls* the builtin as if it was a part
of the builtin itself.

Code for stack trace handling was centralized. We no longer
need traceElement argument in ~every function. Now the stack
trace state is kept solely in the interpreter.


sbarzowski:
Requires https://github.com/google/jsonnet/pull/861.

PR #478, opened 2020-11-23
Support creating VMs from multiple threads in C-bindings

eagletmt:
Since all VM instances are managed by a global map, read/write of the
global map must be protected by a mutex.

### DIFF
--- a/c-bindings/handles.go
+++ b/c-bindings/handles.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"sync"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


sbarzowski:
Good catch, but it's not just a matter of VM handles, but also JsonValues used with native functions. So, this change might have some noticeable performance implications. The overhead for that is pretty big anyway, so probably not a big deal.

sbarzowski:
I thought about it for a moment and I'm pretty sure it won't be noticeable performance-wise in any reasonable scenario.

PR #481, opened 2020-11-24
Add format APIs to C bindings

eagletmt:
I've added `formatOptions` field to `vm` struct to keep API compatibility with the C++ version.


sbarzowski:
Awesome contribution. Thank you!

PR #482, opened 2020-11-24
Remove dependency on cpp-jsonnet submodule from C bindings

eagletmt:
This change allows us to build libgojsonnet from tarball downloaded from
GitHub releases.

### DIFF
--- a/c-bindings/BUILD.bazel
+++ b/c-bindings/BUILD.bazel
@@ -17,12 +17,9 @@ go_library(
         "internal.h",
         "libjsonnet.cpp",
     ],
-    cdeps = [
-        "@cpp_jsonnet//include:libjsonnet",
-    ],
     cgo = True,
-    copts = ["-Wall -Icpp-jsonnet/include"],  # keep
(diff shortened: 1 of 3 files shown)


sbarzowski:
We can probably go forward with this, but I don't like that we can now accidentally diverge from the original headers without noticing it.

Maybe we could keep a copy of the header in the repo and use it for compiling libgojsonnet. We could then check if it's the same as in cpp-jsonnet in tests.sh (which depends on cpp-jsonnet anyway). Let me think about this.

johnbartholomew:
Four and a half years later, this pull request still seems useful to me.

Looking at the history of libjsonnet.h in the C++ repository, the declarations being copied here last changed in 2014: https://github.com/google/jsonnet/commit/2c5fd8716b2719e63b870503b0685cde0e0ddf84#diff-ee89e9022b7fdd09da47ce82a4b9b1fd6901967ffbbd8d1a30249a18414b715f

It would be nice to have an automated check in the test suite that the copied declarations don't diverge... but given the rate of change here it doesn't seem like a high urgency problem.

I've rebased the change onto master.

PR #486, opened 2020-12-12
Fix builds filter ro re-enable darwin builds

lucendio:
I was not able to find any evidences of darwin builds being disabled intentionally. Considering
the [documentation](https://goreleaser.com/customization/build/) I suspect that this was by
mistake. Currently, each list item is applied individually (OR). If I'm wrong, please discard this
PR and share your reasoning.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -36,7 +36,7 @@ builds:
       - arm64
     ignore:
       - goos: darwin
-      - goarch: 386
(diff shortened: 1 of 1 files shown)


sbarzowski:
Good catch, thanks!

joe-getcouragenow:
any chance and darwin amd64 release can happen ?

PR #492, opened 2020-12-30
Add std. manifestJsonMinified

### LINKED ISSUE #471: Feature request: standard library function to produce "compressed" / "minimal whitespace" json
We have a place in our code, where we want to emit a stringified JSON value. Suppose the intended output looks something like:

```json
{
   "foo": "bar",
   "myJSONValue": "{\"hello\":\"world\",\"one\":1}"
}
```

When we use `std.manifestJsonEx(obj, '')`, the stringified JSON contains newlines, and spaces between key-value pairs, e.g. 

```
{
   "foo": "bar",
   "myJSONValue": "{\n\"hello\": \"world\",\n\"one\": 1\n}"
}
```

Of course, our real application of this is serializing an object with many key-value pairs, and the value is intended to be consumed by a machine.

We considered using string replacement to trim some of the whitespace away, but that seems like a brittle solution.

Would you be open to adding a new function like `manifestJsonCompressed` or `manifestJsonMinimized` which would result in the former output, without unnecessary whitespace?

hanyucui:
This resolves #471

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -1211,13 +1211,31 @@ func jsonEncode(v interface{}) (string, error) {
 // These should ideally be unified
 // For backwards compatibility reasons, we are manually marshalling to json so we can control formatting
 // In the future, it might be apt to use a library [pretty-printing] function
-func builtinManifestJSONEx(i *interpreter, obj, indent value) (value, error) {
-	vindent, err := i.getString(indent)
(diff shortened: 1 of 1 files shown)


hanyucui:
@sbarzowski Let me know if I missed anything.

sbarzowski:
LG, thanks!

PR #494, opened 2021-01-11
sketch: allow initialization of external and TLA code variables from an AST node

gotwarlost:
This allows configuration of a VM with an external code variable that has been pre-parsed into an AST node.
of converting the object to an `ast.Node` is incurred only once by the caller as opposed to having this

If the implementation looks acceptable, I'll add the tests.


gotwarlost:
commited with the wrong username. will reopen

PR #495, opened 2021-01-11
sketch: allow initialization of external and TLA code variables from an AST node, fixes #493

### LINKED ISSUE #493: Feature request: Allow an ext code var to be initialized using an ast.Node
I have a use-case where a piece of jsonnet code that is costly to eval returns an object and I want to use this object for subsequent invocations in separate VMs as an external code var. This object is big and not everything that is subsequently eval-ed needs to use it.

Each eval runs in an isolated VM partly because of concurrent processing of components and partly because every invocation has a different set of vars which makes the import cache useless anyway. That is, instead of re-using VMs in complex ways taking concurrency into account etc., my code just creates a fresh VM every time.

What would be really helpful is if I could turn the initial object to an `ast.Node` using `SnippetToAST` and initialize it as `vm.ExtCodeNode(name, node)` or something. 

Passing in the serialized version of the object negates the gains I've made in the preprocessing in that the extVar is eagerly turned into a node every time even if the component code doesn't happen to require the value of the var. And being a giant object, it happens to be costly to parse.

I hacked together some crappy code to implement `ExtCodeNode` in my local vendor tree and verified that what I'm describing is indeed the bottleneck that makes everything slow.

Other ideas also welcome.

gotwarlost:
This allows configuration of a VM with an external code variable that has been pre-parsed into an AST node.
of converting the object to an ast.Node is incurred only once by the caller as opposed to having this

If this looks acceptable, I'll add some tests.

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -46,18 +46,23 @@ type VM struct {
 	importCache    *importCache
 }
 
+// extKind indicates the kind of external variable that is being initialized for the VM
+type extKind int
+
+const (
+	extKindString extKind = iota // a simple string
+	extKindCode                  // a code snippet represented as a string
+	extKindNode                  // an ast.Node that is passed in
+)
+
 // External variable or top level argument provided before execution
 type vmExt struct {
-	// jsonnet code to evaluate or string to pass
+	// the kind of external variable that is specified.
+	kind extKind
+	// jsonnet code to evaluate (kind=extKindCode) or string to pass (kind=extKindString)
 	value string
-	// isCode determines whether it should be evaluated as jsonnet code or
-	// treated as string.
-	isCode bool
-	// an ext or TLS code variable can also be initialized from an AST node
+	// the specified node for kind=extKindNode
 	node ast.Node
-	// isNode determines if a code variable has been initialized using an AST node
-	// as opposed to a code string.
-	isNode bool
 }
 
 type vmExtMap map[string]vmExt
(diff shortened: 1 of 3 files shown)


vm.go
```diff
@@ -53,6 +53,11 @@ type vmExt struct {
 	// isCode determines whether it should be evaluated as jsonnet code or
 	// treated as string.
 	isCode bool
+	// an ext or TLS code variable can also be initialized from an AST node
+	node ast.Node
+	// isNode determines if a code variable has been initialized using an AST node
+	// as opposed to a code string.
+	isNode bool
```
sbarzowski, line 60: Let's merge isCode and isNode into an enum and replace nested ifs with a single switch.

sbarzowski (commented):
LG. One small nit. Other than that it's only missing the tests.

gotwarlost:
In which file do you want me to add the tests? would it be in `jsonnet_test.go`?

sbarzowski:
> In which file do you want me to add the tests? would it be in `jsonnet_test.go`?

Yes, that's a good location. 

It doesn't have to be too extensive. It's enough to check that the API works with one example for each function. (Unless ofc you can think of some important properties to check in addition to that.)

gotwarlost:
Turns out I need to get this repo approved by our company's open-source committee for upstream contributions (I thought I had already done this in the past). I'm closing this PR - will create a new one once the approvals are done in a few days.

PR #497, opened 2021-01-13
Distinguish the Go implementation

hanyucui:


hanyucui:
Depends on google/jsonnet#882

sbarzowski:
Can you run `./update_cpp_jsonnet.sh` again? We're updating cpp-jsonnet anyway and an important bugfix was merged today, so it would be good to include it.

sbarzowski:
Line numbers changed in in stdlib, so it's complaining about error messages not matching. It can be fixed with `go test -update`

sbarzowski:
Ah, and the cpp-jsonnet version also needs to be updated in Bazel manifests:
https://github.com/google/go-jsonnet/blob/v0.17.0/bazel/repositories.bzl#L32

Note that commit numbers also work for archive, so you can just replace 0.17.0 with a specific commit there as well (and in strip_prefix).

Sorry about extra hassle. This should have been mentioned in the README or maybe even automated with update_cpp_jsonnet.

hanyucui:
> cpp-jsonnet

Does the sha256 need to be updated?

sbarzowski:
Thanks a lot for doing this!

Issue #496, opened 2021-01-12
Optional plus sign makes writing Jsonnet code error prone


curusarn:
Consider following example:
```
❯ cat example.jsonnet    
[
  { key1: "something", key2: "something" }
  { key1: "something_else", key2: "something_else" },
]

❯ jsonnet example.jsonnet
[
   {
      "key1": "something_else",
      "key2": "something_else"
   }
]
```

In the example above, there is a missing comma which results in the two objects being merged together. Since the objects have the same fields, the first object simply disappears from the output. Errors like this one are very hard to spot.

I'm wondering why is the "plus sign" / "merge operator" optional. Requiring the operator would make it obvious if the programmer wants to merge objects or if they just forgot a comma. Is there a recommended way to avoid errors like this one? Am I missing something?

Follow up question:
Would it make sense to have an option for `jsonnetfmt` that produces Jsonnet code that includes the optional "merge operators"?

seh:
Pertinent: #233.

sbarzowski:
> I'm wondering why is the "plus sign" / "merge operator" optional. Requiring the operator would make it obvious if the programmer wants to merge objects or if they just forgot a comma. Is there a recommended way to avoid errors like this one? Am I missing something?

The implicit syntax is quite natural for patterns similar to traditional inheritance:
```
local foo = some_template {
    to_be_overriden: bar;
}
```

That said, this convenience is probably not worth the bugs it has caused in other contexts. That ship has sailed, though. The best we can do is to have linter warnings (#233) or other mechanism to minimize the impact.

> Would it make sense to have an option for `jsonnetfmt` that produces Jsonnet code that includes the optional "merge operators"?

Yes, that would be a good thing. Maybe even better than linter support, because it applies the fixes automatically.

PaulRudin:
Another possibility would be to have a command line option to error on implicit +. That way backwards compatibility works, but if people want to enforce it they can.

sbarzowski:
Generally we want to avoid creating language dialects, as much as possible to keep the community together. Having a switch like that would create a lot of problems e.g. with 3rd party libraries which may or may not be compatible.

What we might do one day is pragma statements which change things locally, only for a specific file. This way we can migrate one file at a time. This still introduces a lot of complexity, because all pragma combinations will need to be supported indefinitely in all interpreter implementations. There is still plenty we can do without doing that, so I wouldn't expect pragmas anytime soon. 

And in the meantime, linter warnings are a good solution, once we add them.

[...]

PR #506, opened 2021-01-22
Make jsonnetfmt add plus objects instead of removing them

curusarn:
Implements: https://github.com/google/go-jsonnet/issues/496

- I reused the `fix_plus_sign.go` file instead of removing it and creating a new one. Let me know if you would like rename it.
- I don't add new `Fodder` nor interact with existing `Fodder` in any way. I think that is correct but that's for someone more knowledgeable to say.
- I'm not sure if the comments are correct so please check them.

### DIFF
--- a/cmd/jsonnetfmt/cmd.go
+++ b/cmd/jsonnetfmt/cmd.go
@@ -58,6 +58,8 @@ func usage(o io.Writer) {
 	fmt.Fprintln(o, "  --[no-]pad-objects         { x: 1, y: 2 } instead of {x: 1, y: 2}")
 	fmt.Fprintln(o, "                             (on by default)")
 	fmt.Fprintln(o, "  --[no-]sort-imports        Sorting of imports (on by default)")
+	fmt.Fprintln(o, "  --[no-]use-implicit-plus   Remove plus signs where they are not required")
+	fmt.Fprintln(o, "                             (on by default)")
 	fmt.Fprintln(o, "  --version                  Print version")
 	fmt.Fprintln(o)
 	fmt.Fprintln(o, "In all cases:")
@@ -161,6 +163,10 @@ func processArgs(givenArgs []string, config *config, vm *jsonnet.VM) (processArg
 			default:
 				return processArgsStatusFailure, fmt.Errorf("invalid --comment-style value: %s", str)
 			}
+		} else if arg == "--use-implicit-plus" {
+			config.options.UseImplicitPlus = true
+		} else if arg == "--no-use-implicit-plus" {
+			config.options.UseImplicitPlus = false
 		} else if arg == "--pretty-field-names" {
 			config.options.PrettyFieldNames = true
 		} else if arg == "--no-pretty-field-names" {
--- a/internal/formatter/add_plus_object.go
+++ b/internal/formatter/add_plus_object.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2019 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package formatter
+
+import (
+	"github.com/google/go-jsonnet/ast"
+	"github.com/google/go-jsonnet/internal/pass"
+)
+
+// AddPlusObject is a formatter pass that replaces e {} with e + {}.
+type AddPlusObject struct {
+	pass.Base
+}
+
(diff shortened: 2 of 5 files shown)


sbarzowski:
This modification should be behind a new option flag. You can leave the old FixPlusObject as it was and copy the new version with a different name. Perhaps you can also rename the old one, so they're both more specific. 

Regarding fodder, I think we're good, but we should add a test for that anyway. A good test would be to have implicit + with as many comments as possible. E.g. `/* a */ { /* b */ foo: /* c */ 1 /* d */ } /* e */ { /* f */ foo: /* g */ 2 /* h */ } /* i */`. If the output doesn't lose or move any comments weirdly, we're good.

sbarzowski:
Ah, it seems that for formatter, we're depending completely on cpp-jsonnet test suite. Hmmm... Let's just test it manually for now and I'll think what we can do about it.

curusarn:
> This modification should be behind a new option flag. 

Ah, I see. I got an impression that @seh suggested that adding plus objects could be the new default behavior.

seh:
I'd prefer that that become the default, but I defer to @sbarzowski to decide whether we can afford the change in the golden test files.

sbarzowski:
It might become the default later, but let's not get ahead of ourselves :-).

curusarn:
I suppose that no matter what the default is we still want to have an option for the other behavior. So I will change the PR to add an option for the new behavior.

curusarn:
Do you have an idea for how the command-line option should be named? Perhaps `--[no-]use-implicit-plus` or
`--[no-]forbid-implicit-plus`?

sbarzowski:
I like `--[no-]use-implicit-plus`.

curusarn:
@sbarzowski Can you point me to the test files for the formatter? I found testdata and I assume it is used by linter, formatter, and parser. But I don't know how exactly.

curusarn:
If there are some docs about it just point me there. I don't want to waste your time.

sbarzowski:
The formatter is tested only using cpp-jsonnet test suite. We are interested in "cmd" tests (b/c we're not changing the output with default settings). Here they are: https://github.com/google/jsonnet/blob/v0.17.0/test_cmd/run_cmd_tests.sh. If you run into any trouble feel free to leave it out and I'll do it myself.

curusarn:
Here is a diff with test for `--no-use-implicit-plus`:
```diff
diff --git a/test_cmd/fmt_no_use_implicit_plus.golden.stdout.golang b/test_cmd/fmt_no_use_implicit_plus.golden.stdout.golang
new file mode 100644
index 0000000..3d46393
--- /dev/null
+++ b/test_cmd/fmt_no_use_implicit_plus.golden.stdout.golang
@@ -0,0 +1,5 @@
+[
+  /* a */
+  { /* b */ foo: /* c */ 1 /* d */ } + /* e */
+  { /* f */ foo: /* g */ 2 /* h */ }, /* i */
+]
diff --git a/test_cmd/fmt_no_use_implicit_plus.stdin b/test_cmd/fmt_no_use_implicit_plus.stdin
new file mode 100644
index 0000000..0332273
--- /dev/null
+++ b/test_cmd/fmt_no_use_implicit_plus.stdin
@@ -0,0 +1,4 @@
+[
+  /* a */ { /* b */ foo: /* c */ 1 /* d */ } /* e */
+  { /* f */ foo: /* g */ 2 /* h */ } /* i */
+]
diff --git a/test_cmd/run_cmd_tests.sh b/test_cmd/run_cmd_tests.sh
index 9d9086f..2703ec5 100755
--- a/test_cmd/run_cmd_tests.sh
+++ b/test_cmd/run_cmd_tests.sh
@@ -128,6 +128,7 @@ if do_fmt_test "fmt_out" 0 -e "{ a: 1, b: 2, c: 3 }" -o "out/fmt_out/custom_outp
     check_file "fmt_out" "out/fmt_out/custom_output" "fmt_out.golden.custom_output"
 fi
 do_fmt_test "fmt_double_dash" 0 -e -- -1
+do_fmt_test "fmt_no_use_implicit_plus" 0 --no-use-implicit-plus
 
 if mkdir -p "out/fmt_inplace" && cp "test.jsonnet" "out/fmt_inplace/test.jsonnet"; then
     # Test jsonnetfmt in-place modifications
```

Is this okay? Should I create a separate PR in `google/jsonnet` or can you just apply the diff?

I tried running the tests and I encountered some issues. I didn't figure out how to refresh existing golden files for jsonnetfmt (jsonnetfmt --help doesn't match).

sbarzowski:
It's fine. I'll apply the change on the cpp-jsonnet side. So, it seems the only thing that we're still missing here is the CLA.

curusarn:
I have passed the CLA to my employer's legal department. I will need to wait for them to review it and sign it. Not sure how long that is going to take.

curusarn:
@googlebot I signed it!

curusarn:
@sbarzowski I don't mean to urge you. I'm just pinging you to let you know that everything should be ready.

sbarzowski:
Thank you for the reminder! Generally feel free to ping on issues and especially PRs when they're taking too long.

I can't promise that I will always reply quickly (in the recent weeks I had ~0 free capacity).

PR #508, opened 2021-02-01
allow initialization of external and TLA code variables from an AST node

gotwarlost:
Allow configuration of a VM with an external code variable that has been pre-parsed into an AST node.
of converting the object to an ast.Node is incurred only once by the caller as opposed to having this


sbarzowski (approved):

sbarzowski:
Rebasing is necessary.

gotwarlost:
rebased.

sbarzowski:
Thanks!

PR #509, opened 2021-02-01
Allow reset of ext and TLA vars for VM reuse

gotwarlost:
Currently ext and TLA vars can be set but not reset. This makes
re-using VMs that have different variables impossible.

Add VM methods to be able to reset ext and TLA vars

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -115,7 +115,7 @@ func (vm *VM) TLACode(key string, val string) {
 	// Setting a TLA does not require flushing the cache - see above.
 }
 
-// TLAReset resets all TLSa registered for this VM.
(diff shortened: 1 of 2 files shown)


vm.go
```diff
@@ -109,6 +115,11 @@ func (vm *VM) TLACode(key string, val string) {
 	// Setting a TLA does not require flushing the cache - see above.
 }
 
+// TLAReset resets all TLSa registered for this VM.
```
sbarzowski, line 118: Typo: TLSa -> TLAs

sbarzowski (approved):
Looks good. Thanks!

vm.go
gotwarlost: fixed

Issue #394, opened 2020-03-19
std.trace should not hardcode stderr


sbarzowski:
When library interface is used we should not assume that `stderr` is the right target for `std.trace` messages. Instead we should provide some API to define where it should go. Probably a Writer interface here in go-jsonnet, but I'm not sure about C api.

sparkprime:
In C it would need to be a callback like the others, which would be called with the string as a param.  It should be fairly easy compared to some of the other C/go interfaces!

hanyucui:
@sbarzowski Want to give this a try. Hopefully, I can start in a week or two. I imagine I will need to first learn the existing library interfaces. Let me know if you have any advice.

sbarzowski:
Cool!

We can have an io.Writer in VM, which can be set with some method. This can then be used in the Fprintf instead of os.Stderr in std.trace implementation (in builtins.go). That should be enough to support it with Go API.

To support it in C API (which can be a separate PR, and doesn't necessarily need to be done at the same time), we need to figure out the interface for the callback or something – we don't have a ready io.Writer in C. 

One way is to have a callback function which basically mirrors io.Writer and then have a custom loop to output the data through such callback. We can still use sprintf to prepare the message. Maybe you can come up with something better.

hanyucui:
@sbarzowski For the C API, it seems the suggestion from @sparkprime is to pass the message string to C in a callback rather than accepting a `FILE*` and writing to it directly, right?

sbarzowski:
Yeah, that would be better. The reason is that we're trying to make the API OS agnostic. AFAIK it's not possible to pass your custom thing as a FILE* in C. So with `FILE*` t would need to refer to some actual file descriptor. Please, correct me if I'm wrong. Go `Writer` doesn't have this problem – it can be anything consuming bytes.

hanyucui:
Makes sense. Let me give it a try.

PR #514, opened 2021-02-08
Allow setting a different output stream than os.Stderr for std.trace

hanyucui:
Allow setting a different output stream than `os.Stderr` for `std.trace`.
This is requested by #394.

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -24,7 +24,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
-	"os"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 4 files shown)


jsonnet_test.go
```diff
@@ -327,3 +328,23 @@ func TestTLATypes(t *testing.T) {
...
 	assertVarOutput(t, jsonStr)
 }
+
+func TestSetTraceOut(t *testing.T) {
+	traceOut := &strings.Builder{}
+	vm := MakeVM()
+	vm.SetTraceOut(traceOut)
+
+	const msg = "Some Trace Message"
+	expected := fmt.Sprintf("TRACE: :1 %s", msg)
```
hanyucui, line 338: Not sure why the file name "blah.jsonnet" does not show up in the message.

builtins.go
```diff
@@ -251,7 +250,7 @@ func builtinTrace(i *interpreter, x value, y value) (value, error) {
 	filename := trace.loc.FileName
```
sbarzowski, line 250: We should probably use trace.loc.File.DiagnosticFilename. The difference is that trace.loc.Filename should only contain filenames which are valid importer paths while DiagnosticFilename may have dummy strings (like the one provided in AnonymousSnippet).

jsonnet_test.go
sbarzowski: I've added a comment above on how to fix this.

sbarzowski (approved):

sbarzowski:
Thank you!

Issue #515, opened 2021-02-15
std.manifestJsonEx breaking change in master


squat:
Hi :wave: In previos versions of go-jsonnet, up to and including v0.17.0, `std.manifestJsonEx` accepted an array for the `value` parameter. As of master, passing an array causes jsonnet to throw an error a la:

```
RUNTIME ERROR: Unexpected type array, expected object
```

This seems like a breaking change in the standard library, however I could not find any references to this in the commits between the v0.17.0 tag and tip of master or any release notes. Is this known and expected behavior? If so, it would be great to have this documented somewhere so that users can expect to migrate their code.

squat:
I bisected the changes and it seems that 7d3bda3911b71dac3d90607b5ec39a12f057640a introduces this behavior

sbarzowski:
This is a serious bug. I should have noticed this and I'm a bit surprised that our tests didn't catch this. Many thanks for reporting this! 

The fix is probably trivial. It looks like it's enough to pass the value from `arguments[0]` directly to `aux` instead of forcing it to be an object.

squat:
yes, that seems like an easy workaround :+1: happy to make a PR and add a test

sbarzowski:
Great! Please go ahead.

PR #516, opened 2021-02-15
Fix manifest json ex regression

squat:
This PR fixes a regression in std.manifestJsonEx that caused the
standard library function to error when the given value was an array.
It also adds a test case to prevent regressions in
std.manifestJsonEx accepting arrays as values.

fixes: #515 

Implements fix suggested by @sbarzowski in https://github.com/google/go-jsonnet/issues/515#issuecomment-779123853

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -1211,10 +1211,7 @@ func jsonEncode(v interface{}) (string, error) {
 // For backwards compatibility reasons, we are manually marshalling to json so we can control formatting
 // In the future, it might be apt to use a library [pretty-printing] function
 func builtinManifestJSONEx(i *interpreter, arguments []value) (value, error) {
-	obj, err := i.getObject(arguments[0])
-	if err != nil {
-		return nil, err
-	}
+	val := arguments[0]
 
 	vindent, err := i.getString(arguments[1])
 	if err != nil {
@@ -1315,7 +1312,7 @@ func builtinManifestJSONEx(i *interpreter, arguments []value) (value, error) {
 		}
 	}
 
-	finalString, err := aux(obj, path, "")
+	finalString, err := aux(val, path, "")
 	if err != nil {
 		return nil, err
 	}
--- a/testdata/builtinManifestJsonEx.jsonnet
+++ b/testdata/builtinManifestJsonEx.jsonnet
@@ -1,4 +1,4 @@
-local a = {
+local object = {
   foo: 'bar',
   bar: self.foo,
   baz: 1,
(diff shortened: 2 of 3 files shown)


testdata/builtinManifestJsonEx.jsonnet
```diff
@@ -1,22 +1,46 @@
...
-    baz: 1,
-    bazel: 1.42,
-    boom: -1,
-    bim: false,
-    bam: true,
-    blamo: {
-        cereal: [
-            "<>& fizbuzz",
-        ],
+  foo: 'bar',
```
squat, line 2: reformatting happened automatically when saving with jsonnetfmt hook, hope that's ok. Let me know if I should revert

testdata/builtinManifestJsonEx.jsonnet
```diff
@@ -1,22 +1,46 @@
...
+      {
+        name: 'chocolate',
+      },
+    ],
+  },
+];
+
+{
+  a: std.manifestJsonEx(a, '  '),
+  b: std.manifestJsonEx(b, '  '),
```
sbarzowski, line 45: Let's also add examples of other data types, namely: null, string, number, bool. They all should work.

testdata/builtinManifestJsonEx.jsonnet
sbarzowski: It's fine. We should probably keep them all formatted (besides the ones intended to test the formatter).

builtins.go
```diff
@@ -1211,11 +1211,6 @@ func jsonEncode(v interface{}) (string, error) {
 // For backwards compatibility reasons, we are manually marshalling to json so we can control formatting
 // In the future, it might be apt to use a library [pretty-printing] function
 func builtinManifestJSONEx(i *interpreter, arguments []value) (value, error) {
-	obj, err := i.getObject(arguments[0])
```
sbarzowski, line 1214: Nit: Let's keep it named. `value := arguments[0]`. This way it's more clear what the args are.

sbarzowski (approved):

builtins.go
squat: calling it `value` will shadow the type with the same name, which breaks the funcs that declare parameters of that type inside of the scope of `buildinManifestJSONEx`.
How about `val`?

testdata/builtinManifestJsonEx.jsonnet
squat: sounds great :+1:

builtins.go
sbarzowski: Good point. `val` sounds good.

PR #529, opened 2021-04-23
Prototype: Rudimentary stack profiler

igorwwwwwwwwwwwwwwwwwwww:
We have a relatively large jsonnet code base ([source](https://gitlab.com/gitlab-com/runbooks/-/tree/master/dashboards)), and I was trying to profile slow jsonnet invocations.

The golang-level profile via `JSONNET_CPU_PROFILE` wasn't very helpful in this regard. And tracing individual calls by adding `std.trace` was quite combersome.

I prototyped a little sampling stack profiler. It is not very sophisticated. It samples stacks for 1% of calls to `EvalInCleanEnv`. Those stacks can then be visualised [as a flamegraph](https://github.com/brendangregg/FlameGraph).

Usage:

```
JSONNET_STACK_PROFILE=stacks.out jsonnet main.jsonnet
cat stacks.out | flamegraph.pl --hash > flamegraph.svg
```

There's plenty of room for improvement:

- Making the code nicer
- Including a time component (since we currently assume each eval takes a fixed amount of time)
- Buffering stacks in memory with their counts and writing out aggregated counts in the end

That said, the results I got from it so far look quite usable.

It's unlikely that I'll have much time to work on this and develop it into a fully-fledged solution.

I wanted to share what I have in case it's useful for others, or in case someone feels inspired to take the idea and develop it further.

### DIFF
--- a/cmd/jsonnet/cmd.go
+++ b/cmd/jsonnet/cmd.go
@@ -397,6 +397,9 @@ func main() {
 	cmd.StartCPUProfile()
 	defer cmd.StopCPUProfile()
 
+	jsonnet.StartStackProfile()
+	defer jsonnet.StopStackProfile()
+
 	vm := jsonnet.MakeVM()
 	vm.ErrorFormatter.SetColorFormatter(color.New(color.FgRed).Fprintf)
 
--- a/interpreter.go
+++ b/interpreter.go
@@ -17,13 +17,18 @@ limitations under the License.
 package jsonnet
 
 import (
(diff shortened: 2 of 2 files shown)


sbarzowski:
This is very interesting. I was thinking about doing something like that myself, but didn't have much time recently.

I'd like to refactor it slightly (eliminate globals, extract all profiling logic to a separate function, so that in the normal flow it's just one check and one function call). Otherwise it looks good.

PR #532, opened 2021-05-14
Fix strip definition

isarrigiannis:
Add missing first parameter: the string to be stripped


sbarzowski:
Ah. Could you add the appropriate regression test to this file: https://github.com/google/go-jsonnet/blob/master/testdata/stdlib_smoke_test.jsonnet? See the comment on top for instructions.

To regenerate the output, use `go test ./... -update` in the root directory. Then inspect the diff to make sure there is no unexpected output (e.g. lines in the linter output).

fenech:
@googlebot I fixed it.

fenech:
@googlebot I consent.

sbarzowski:
Yes, thank you!

PR #537, opened 2021-05-26
linter: correct the return type for member

squat:
This commit corrects return type of the `member` function from the
standard library.

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>


squat:
It seems there were some flakes with the am64 and ppc travis tests

sbarzowski:
Thank you!

PR #548, opened 2021-07-11
Fix jsonnet-lint erorr when linting importing files and imported files at the same time: #544

### LINKED ISSUE #544: jsonnet-lint panics on an import that imports the same file
One of our files had an unused variable that was set to an import of the same file.

```
$ cat tmp/foo.libsonnet
local foo = import 'foo.libsonnet';
{}

$ jsonnet-lint tmp/foo.libsonnet
tmp/foo.libsonnet:1:7-35 Unused variable: foo

local foo = import 'foo.libsonnet';


panic: Bug - placeholder for a dependent node cannot be noType

goroutine 1 [running]:
github.com/google/go-jsonnet/linter/internal/types.(*typeGraph).getExprPlaceholder(...)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/build_graph.go:26
github.com/google/go-jsonnet/linter/internal/types.calcTP(0x1200cc0, 0xc00010a100, 0xc00005f440, 0xc0001246c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/build_graph.go:217 +0x3696
github.com/google/go-jsonnet/linter/internal/types.prepareTPWithPlaceholder(0x1200cc0, 0xc00010a100, 0xc00005f440, 0xc0001246c0, 0x72)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/build_graph.go:86 +0x179
github.com/google/go-jsonnet/linter/internal/types.prepareTPWithPlaceholder(0x1200fc0, 0xc000104280, 0xc00005f440, 0xc0001246c0, 0x71)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/build_graph.go:62 +0xb2b
github.com/google/go-jsonnet/linter/internal/types.(*typeGraph).addRoots(0xc0001246c0, 0xc0001df698, 0xc0001df5d8)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/build_graph.go:96 +0x3ee
github.com/google/go-jsonnet/linter/internal/types.Check(0x1200fc0, 0xc000104280, 0xc0001df698, 0xc0001df5d8, 0xc00004a7e0, 0xc0001df5c0)
        /Users/kanantheswaran/go/go-jsonnet/linter/internal/types/check.go:139 +0xb4
github.com/google/go-jsonnet/linter.lint(0xc00005c0a0, 0x1200fc0, 0xc000104280, 0x7ffeefbff02d, 0x11, 0xc0001dfdb0)
        /Users/kanantheswaran/go/go-jsonnet/linter/linter.go:76 +0x8a4
github.com/google/go-jsonnet/linter.LintSnippet(0xc00005c0a0, 0x11fd6c0, 0xc00000e020, 0x7ffeefbff02d, 0x11, 0xc00001c360, 0x27, 0xc000104140)
        /Users/kanantheswaran/go/go-jsonnet/linter/linter.go:134 +0x13d
main.main()
        /Users/kanantheswaran/go/go-jsonnet/cmd/jsonnet-lint/cmd.go:187 +0x84a
```

thombashi:
This PR aims to fix #544.

I also encounter #544 (the same case with https://github.com/google/go-jsonnet/issues/544#issuecomment-877622528)

Please let me know if my understanding is wrong.
The problem seems to be caused by XxxtoAST (`ImportAST` and `SnippetToAST`) functions return different Node instances, even with the same arguments.
Specifically, about using the `ast.Node` cache. `ImportAST` function (call `SnippetToAST` function internally) uses `ast.Node` cache, while a direct call of `SnippetToAST` function does not use the cache.
So, the return values are different (e.g. `LocationRange.File`).
This will results missing key at `getExprPlaceholder`.

To fix this problem, I made changes to `SnippetToAST` to use the same cache with `ImportAST`.

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -117,6 +117,15 @@ func (cache *importCache) importData(importedFrom, importedPath string) (content
 	return
 }
 
+func (cache *importCache) snippetToAST(diagnosticFilename ast.DiagnosticFileName, importedFilename, snippet string) (ast.Node, error) {
+	if cachedNode, isCached := cache.astCache[importedFilename]; isCached {
+		return cachedNode, nil
+	}
+	node, err := program.SnippetToAST(diagnosticFilename, importedFilename, snippet)
+	cache.astCache[importedFilename] = node
+	return node, err
+}
+
 func (cache *importCache) importAST(importedFrom, importedPath string) (ast.Node, string, error) {
 	contents, foundAt, err := cache.importData(importedFrom, importedPath)
 	if err != nil {
--- a/jsonnet_test.go
+++ b/jsonnet_test.go
@@ -290,11 +290,11 @@ func assertVarOutput(t *testing.T, jsonStr string) {
 }
 
 func TestExtTypes(t *testing.T) {
-	node, err := SnippetToAST("var.jsonnet", `{ node: 'node' }`)
+	vm := MakeVM()
+	node, err := vm.SnippetToAST("var.jsonnet", `{ node: 'node' }`)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
-	vm := MakeVM()
 	vm.ExtVar("var", "var")
 	vm.ExtCode("code", `{ code: 'code'}`)
 	vm.ExtNode("node", node)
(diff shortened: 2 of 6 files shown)


sbarzowski:
Thanks for sending this. I think your diagnosis is correct, but the solution causes other problems. I think that now if you use EvaluateAnonymousSnippet multiple times, you'll actually get the first one evaluated every time. The very idea behind the `Snippet` variants was that it can be used outside of cachable paths. It is supposed to return a new thing each time. 

So to fix this I would go a slightly different route. (a) Figure out why returning different root nodes causes issues in the first place. Intuitively it should be slower, but still work. (b) Change the API of the linter to take the filenames and use `importAST` instead of hacking around the snippets (this is a breaking change, but it's still ok for the linter).

PR #550, opened 2021-07-12
Fix gorelaser configs to properly build darwin packages

thombashi:
This PR aims to add packages for `macOS` at releases.
For now, binary packages for macOS are missing at the release page even though `goreleaser` settings include macOS at `goos`.
This is because ignore config of `goreleaser` are incorrect.
In the current config, all of the macOS(`darwin`) will be ignored.


sbarzowski:
Thank you!

PR #555, opened 2021-08-08
Fix linter hanging in certain situations.

### LINKED ISSUE #541: jsonnet-lint hangs under certain conditions
OK, I'm down to one file that jsonnet-lint refuses to process. It basically hangs and ultimately gets killed by what I'm assuming is an OOM killer on my Mac. The weird thing is that _sometimes_ it will return just fine in sub-second.

That file has all sorts of proprietary stuff in it so I've tried my best to create a repro test case but you'll have to tell me if it worked :)

https://gist.github.com/gotwarlost/756e5d55f14d8c8fe28e8ec71e40d1d9

* Create an empty directory and download `create-lint-problems.go` and `bash-loop.sh`  from the gist above to it
* run `bash-loop.sh` - it should show you processing times increase as the number of keys in the `caller.jsonnet` map increases and will eventually reach a point where jsonnet-lint doesn't return.

I've no idea what's going but it feels like some sort of quadratic/ exponential processing, or some lock contention. As I mentioned, the weird part is that every so often it will return just fine.

sbarzowski:
We had some accidental aliasing due to shallow copy instead of deep copy.

Fixes #541.

### DIFF
--- a/linter/internal/types/build_graph.go
+++ b/linter/internal/types/build_graph.go
@@ -192,6 +192,9 @@ func calcTP(node ast.Node, varAt map[ast.Node]*common.Variable, g *typeGraph) ty
 	case *ast.Error:
 		return concreteTP(voidTypeDesc())
 	case *ast.Index:
+		// TODO(sbarzowski) It appears we create a separate index node each time.
+		// Perhaps we could de-duplicate (cache) index placeholders to save processing
+		// later?
 		switch index := node.Index.(type) {
 		case *ast.LiteralString:
 			return tpIndex(knownObjectIndex(g.getExprPlaceholder(node.Target), index.Value))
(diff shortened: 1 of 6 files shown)


sparkprime:
@sbarzowski this looks useful, shall I try to merge it?

sbarzowski:
Yeah, I need to revive this. I think the travis failure was the primary reason I didn't.

rraj-splunk:
@sbarzowski : Any specific reason why this was never merged ?

davidreuss:
We have a rather large code base where this patch also allows us to lint it rather than having it hanging. It certainly would be great it if could get merged, so we don't have to carry the complexity of a custom build ourselves 🙏

Issue #295, opened 2019-07-20
Creating a wrapper for node js support


limitlessloop:
Hi,

I'm trying to create a wrapper for a node js support using the `go-jsonnet` version of jsonnet rather than the the C++ version. I'm trying to use `gopherjs` to transpile `go-jsonnet` which can be used for node.js applications. I'm hoping by doing this, it will prove faster than the C++ version using emscripten and 🤞will also support imports.

I have an example go app setup following some advice from this [article](https://medium.com/@kentquirk/how-to-use-gopherjs-to-turn-go-code-into-a-javascript-library-1e947703db7a)

```go
package main

//go:generate gopherjs build --minify

// This is an experiment to see if gopherjs can reasonably generate js code from go source
// so that we can have a single-source solution for keys and addresses.
// Use "go generate" to build this.

import (
	"crypto/sha256"

	"github.com/gopherjs/gopherjs/js"
)

func main() {
	js.Module.Get("exports").Set("hashit", hashit)
}

func hashit(s string) []byte {
	h := sha256.New()
	h.Write([]byte(s))
	return h.Sum(nil)
}
```
However I'm unsure what I need to reference from jsonnet after following these instructions:

```bash
git clone github.com/google/go-jsonnet
cd go-jsonnet
go build ./cmd/jsonnet
```
Is there any chance anyone can point me in the right direction?

sbarzowski:
This may be very helpful. We wanted to do that to use on the website (emscripten version was often problematic). 

I'm assuming by "what I need to reference from Jsonnet" you mean the instructions for using the Jsonnet interpreter s a library to run some Jsonnet programs. So, to run Jsonnet program you need to create the [VM](https://godoc.org/github.com/google/go-jsonnet#VM) using [MakeVM](https://godoc.org/github.com/google/go-jsonnet#MakeVM), then optionally configure it using [ExtVar](https://godoc.org/github.com/google/go-jsonnet#VM.ExtVar), [TlaVar](https://godoc.org/github.com/google/go-jsonnet#VM.TLAVar) etc. and then you can run some Jsonnet program (passed as string) using [EvaluateSnippet](https://godoc.org/github.com/google/go-jsonnet#VM.EvaluateSnippet).

Let me know if you have any other questions.

limitlessloop:
[...]

PR #561, opened 2021-09-03
Add wasm build

sparkprime:
@sbarzowski @buremba please take a look

This is inspired by the code @buremba posted in #295

I have another modification in the C++ repo that makes the website use this instead of the emscripten build and it seems to work...


sparkprime:
The website change is https://github.com/google/jsonnet/pull/940

sparkprime:
We might have to disable the wasm build in tests or something

sbarzowski:
Cool stuff. We need to resolve the build error, though. Perhaps it's enough to add build constraints to `cmd/wasm/main.go`?  Right now it's complaining, because it tries to build stuff for linux and we're asking to import `syscall/js`.

sparkprime:
Yes you have to use GOOS=js GOARCH=wasm to compile a go file that imports syscall/js.

So, clearly we don't want to build it as part of our regular stuff.  But we probably do want to build it as standard just to test for compilation errors (tests would be good too but not sure how to achieve that without a JS test framework which might be overkill).

sparkprime:
Does it need something for bazel too?

sparkprime:
Hmm go1.11.x seems not to support it:
```
+ go build -o libjsonnet.wasm ./cmd/wasm
# github.com/google/go-jsonnet/cmd/wasm
cmd/wasm/main.go:63:48: p[2].JSValue undefined (type js.Value has no field or method JSValue)
cmd/wasm/main.go:67:52: p[3].JSValue undefined (type js.Value has no field or method JSValue)
cmd/wasm/main.go:71:54: p[4].JSValue undefined (type js.Value has no field or method JSValue)
cmd/wasm/main.go:75:52: p[5].JSValue undefined (type js.Value has no field or method JSValue)
cmd/wasm/main.go:79:54: p[6].JSValue undefined (type js.Value has no field or method JSValue)
cmd/wasm/main.go:121:83: undefined: js.Func
cmd/wasm/main.go:122:9: undefined: js.FuncOf
cmd/wasm/main.go:123:41: undefined: js.FuncOf
note: module requires Go 1.13make: *** [Makefile:45: test] Error 2
```

sparkprime:
@sbarzowski Good to go now?

PR #575, opened 2021-12-02
Document installation for go >= 1.17

nicklasfrahm:
Since go `1.17` the installation of binaries with `go get` is deprecated. The README.md was updated to reflect this.

I also highlighted the fact that the version suffix is mandatory. This information can be found by running `go help install`, but I believe explicitly specifying it may reduce the entry barrier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,13 @@ This code is known to work on Go 1.12 and above. We recommend always using the n
 
 ## Installation instructions
 
-```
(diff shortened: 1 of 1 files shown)


nicklasfrahm:
I signed the CLA now, so the CI needs to be triggered again.

sbarzowski:
Thank you!

PR #582, opened 2021-12-22
Update to the latest rules_go and use the new keyword argument when registering toolchains

chrismgrayftsinc:
The newer version of rules_go gets added by other rules and causes a failure when registering toolchains with the wrong keyword argument.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -11,7 +11,7 @@ load(
 
 def jsonnet_go_dependencies():
     go_rules_dependencies()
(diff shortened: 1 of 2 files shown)


chrismgrayftsinc:
This is not yet working on my repo.  No need to look at it yet.

chrismgrayftsinc:
This was the wrong solution to the problem I was having.  It might be good to upgrade anyway, but without changing the keyword argument.  I will close this PR.

PR #584, opened 2022-01-23
Add 'importbin' statement

anguslees:
Add `importbin` statement.  Similar to `importstr` but the result is an array of numbers (all integers 0-255).


linter/internal/types/build_graph.go
```diff
@@ -217,6 +217,8 @@ func calcTP(node ast.Node, varAt map[ast.Node]*common.Variable, g *typeGraph) ty
 		return tpRef(g.getExprPlaceholder(imported))
 	case *ast.ImportStr:
 		return tpRef(stringType)
+	case *ast.ImportBin:
+		return tpRef(anyArrayType)
```
anguslees, line 221: I wasn't confident on this line - is this correct?  Is there a better value to return?

sbarzowski: It's correct, but it would be better to return an array of numbers specifically. We can add another hardcoded type for this (alongside anyArrayType):
* Here you can add a new type id: https://github.com/google/go-jsonnet/blob/74b3a1b66972ebc132fc3f3c2778ebc4b76b931b/linter/internal/types/placeholder.go#L15
* Here you actually define it: https://github.com/google/go-jsonnet/blob/74b3a1b66972ebc132fc3f3c2778ebc4b76b931b/linter/internal/types/graph.go#L131
* Here's another place which could benefit from it: https://github.com/google/go-jsonnet/blob/cb9061e99e6769ccc95c697acd8354370efc113d/linter/internal/types/stdlib.go#L9

sbarzowski:
Good stuff. We need C++ support before documenting this, but it's not a blocker for merging this here.

linter/internal/types/build_graph.go
anguslees: Thanks for the hints.  Added `numberArrayType` and a testcase.

sparkprime:
Strategically, I'm tempted to say we could merge this without C++ support and document that as an explicit limitation of the C++ implementation.  What do you think?

imports.go
```diff
@@ -58,19 +59,33 @@ type Importer interface {
...
+	return *(*string)(unsafe.Pointer(c.data))
+}
+
+func (c Contents) Data() []byte {
 	return *c.data
 }
 
 // MakeContents creates Contents from a string.
 func MakeContents(s string) Contents {
+	data := []byte(s)
```
anguslees, line 79: It suddenly occurred to me that this probably breaks the 'repeating MakeContents(foo) produces the exact same Contents' rule.  Might need to change the Contents representation to store both a []byte pointer and a string pointer, or improve the 'is the same' check.

anguslees: False alarm?  I think there's no problem here, and we're good to proceed.

---

More info: I think we get a different `Contents` even when called with the same string, ie:
```golang
foo := "foo"
a, b := MakeContents(foo), MakeContents(foo)
a != b
```
... because of the above line of code ... but afaics this is ok.

Afaics the intended semantics ([enforced](https://github.com/google/go-jsonnet/blob/fb6c700f29d1bcc6b0be96f042670094c69a7001/imports.go#L111-L113) by `importCache.importData()`) are that `Importer.Import` returns a copy of the _same_ `Contents` object for the same `foundAt` value -- not a new `Contents` object made from the same string (subtle difference).

ie: Importer.Import() is meant to do this:
```golang
// Good example
func (i *myimporter) ImportData(from, path string) (contents Contents, foundAt string, err error) {
        foundAt = i.resolvePath(from, path)
        if contents = i.cache[foundAt]; contents == nil {
                contents = MakeContents(i.readContents(foundAt))
                // Important! Return (a copy of) the same `Contents` for a given `foundAt`
                i.cache[foundAt] = contents
        }
        return
}
```

Not this:

```golang
// Wrong/bad example
func (i *myimporter) ImportData(from, path string) (contents Contents, foundAt string, err error) {
        foundAt = i.resolvePath(from, path)
        var stringData string
        if stringData, ok = i.dataCache[foundAt]; !ok {
                stringData = i.readContents(foundAt)
                i.dataCache[foundAt] = stringData
        }
        // Wrong! Now we have a 'new' Contents object for the same `foundAt`
        contents = MakeContents(stringData)
        return
}

sbarzowski:
> Strategically, I'm tempted to say we could merge this without C++ support and document that as an explicit limitation of the C++ implementation. What do you think?

Here are the options I see:
(a) Merge this and then implement importbin on C++ side (which shouldn't be very hard).
(b) Deprecate the C++ implementation and don't spend much effort porting features there anymore (while enforcing that everything that works on C++ side, works in go-jsonnet as well). I wouldn't make an exception just for importbin. Either we want feature-compatibility or we drop it.

imports.go
sbarzowski: Yes, exactly. We enforce that it's the same Contents object. 

(It encourages efficient caching by avoiding unnecessary copies and we can quickly compare pointers for equality instead of going over the whole string).

anguslees:
C++ PR: https://github.com/google/jsonnet/pull/976

I note there's a Rust PR already in flight too https://github.com/CertainLach/jrsonnet/pull/77.  The Rust implementation looks very nice.  As a long-term support plan, I vote for dropping the C++ library in favour of the Rust-produced C bindings.  We can discuss which project (go or Rust) produces the 'official' /usr/bin/jsonnet CLI tool.

Issue #570, opened 2021-11-07
go-jsonnet crashes when passing negative index to std.substr


shoaniki:
```
$ jsonnet -e 'std.substr("hello", -1, 2)'
INTERNAL ERROR: (CRASH) runtime error: slice bounds out of range [-1:]
goroutine 1 [running]:
runtime/debug.Stack(0x140001deae8, 0x1001d8c40, 0x14000178318)
	/opt/homebrew/Cellar/go/1.15.6/libexec/src/runtime/debug/stack.go:24 +0x88
github.com/google/go-jsonnet.(*VM).evaluateSnippet.func1(0x140001dfca8)
	/private/tmp/go-jsonnet-20201225-66900-1ic2o5q/go-jsonnet-0.17.0/vm.go:173 +0x50
panic(0x1001d8c40, 0x14000178318)
	/opt/homebrew/Cellar/go/1.15.6/libexec/src/runtime/panic.go:965 +0x14c
github.com/google/go-jsonnet.builtinSubstr(0x1400012a120, 0x1001ee5f8, 0x14000114a20, 0x1001ee670, 0x140001244f8, 0x1001ee670, 0x14000124500, 0x1400012a1e0, 0x3, 0x3, ...)
	/private/tmp/go-jsonnet-20201225-66900-1ic2o5q/go-jsonnet-0.17.0/builtins.go:1026 +0xa78
github.com/google/go-jsonnet.(*ternaryBuiltin).evalCall(0x14000106930, 0x140001149f0, 0x3, 0x3, 0x100490e40, 0x0, 0x0, 0x0, 0x1400012a120, 0x3, ...)
	/private/tmp/go-jsonnet-20201225-66900-1ic2o5q/go-jsonnet-0.17.0/builtins.go:1451 +0x190
github.com/google/go-jsonnet.(*valueFunction).call(0x14000105050, 0x1400012a120, 0x140001149f0, 0x3, 0x3, 0x100490e40, 0x0, 0x0, 0x0, 0x100126b98, ...)
	/private/tmp/go-jsonnet-20201225-66900-1ic2o5q/go-jsonnet-0.17.0/value.go:349 +0xd8
github.com/google/go-jsonnet.(*interpreter).evaluateTailCall(0x1400012a120, 0x14000105050, 0x140001149f0, 0x3, 0x3, 0x100490e40, 0x0, 0x0, 0x0, 0x1, ...)
[...]

PR #597, opened 2022-03-16
fix: no panic with negative index to std.substr

jdockerty:
As mentioned in #570.

The approach here is similar to the [cpp implementation](https://github.com/google/jsonnet/blob/master/core/vm.cpp#L1494) of `jsonnet`, we catch the problem early and inform the user with a sensible message, as opposed to a large panic stacktrace being shown.


sbarzowski:
Thank you!

PR #599, opened 2022-04-05
Bump dependency to fatih/color

erikgb:
We are experiencing issue with checksum mismatch. Attempting to bump dependency to fatih/color to fix this.

````
$ go install github.com/google/go-jsonnet/cmd/jsonnet@latest
go: downloading github.com/google/go-jsonnet v0.18.0
go: downloading github.com/fatih/color v1.10.0
go: downloading sigs.k8s.io/yaml v1.1.0
pkg/mod/github.com/google/go-jsonnet@v0.18.0/cmd/jsonnet/cmd.go:28:2: github.com/fatih/color@v1.10.0: verifying module: checksum mismatch
	downloaded: h1:Tr0pkf8RgNWBadn2nj0GE5+l8S9ABNy19/ElCYCpzGg=
	sum.golang.org: h1:s36xzo75JdqLaaWoiEHk767eHiwo0598uUxyfiPkDsg=
SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
For more information, see 'go help module-auth'.
````

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -22,8 +22,8 @@ def jsonnet_go_dependencies(go_sdk_version = "host"):
     go_repository(
         name = "com_github_fatih_color",
         importpath = "github.com/fatih/color",
-        sum = "h1:s36xzo75JdqLaaWoiEHk767eHiwo0598uUxyfiPkDsg=",
-        version = "v1.10.0",
+        sum = "h1:mRhaKNwANqRgUBGKmnI5ZxEk7QXmjQeCcuYFMX2bfcc=",
+        version = "v1.12.0",
     )
 
     go_repository(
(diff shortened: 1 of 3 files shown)


erikgb:
@sbarzowski Any clues why the build is failing? Seems unrelated to the change. 🤔

sbarzowski:
Thanks!

Well, the new toolchain version must have broken something. It is unrelated to this change.

sparkprime:
Is this still needed since I bumped the x/sys dependency?  Or was this actually a better fix for that problem?

erikgb:
> Is this still needed since I bumped the x/sys dependency? Or was this actually a better fix for that problem?

No, not really. We managed to work around the issue. Want me to close this?

sparkprime:
Ok I think I understand -- the fatih issue is separate to the x/sys one although they were both in your original PR.  I'll accept this then because I don't see any reason not to bump fatih as well.

PR #604, opened 2022-05-03
Fix Bazel setup to allow building from external repos

psalaberria002:
When loading this repo from another Bazel workspace we get the following error.
```
ERROR: Traceback (most recent call last):
        File "/Users/paulsmsm/cognite/infrastructure/WORKSPACE", line 447, column 14, in <toplevel>
                jsonnet_repos()
        File "/Users/paulsmsm/cognite/infrastructure/tools/jsonnet/repos.bzl", line 8, column 28, in jsonnet_repos
                jsonnet_go_dependencies()
        File "/private/var/tmp/_bazel_paulsmsm/d6487c682645eb55f96a41598d44250b/external/jsonnet_go/bazel/deps.bzl", line 14, column 27, in jsonnet_go_dependencies
                go_register_toolchains(version = "host")
        File "/private/var/tmp/_bazel_paulsmsm/d6487c682645eb55f96a41598d44250b/external/io_bazel_rules_go/go/private/sdk.bzl", line 458, column 13, in go_register_toolchains
                fail("go_register_toolchains: version set after go sdk rule declared ({})".format(", ".join([r["name"] for r in sdk_rules])))
Error in fail: go_register_toolchains: version set after go sdk rule declared (go_sdk)
```
If a repo is already using Go they have probably already defined a Go toolchain, so it's not necessary to declare that as a dependency of go-jsonnet. I have moved the toolchain registration to the WORKSPACE in order to fix this.

I have also executed `bazel run gazelle` which made lots of changes to BUILD files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ bazel run //:gazelle
 To use `go-jsonnet` from a different Bazel workspace, add the repository
 and its dependencies to your WORKSPACE file.
 
-You will also need to register a Go toolchain.
+You will also need to register a Go toolchain (unless you already have one).
 
 It should look like this:
 
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Additionally if any files were moved around, see the section [Keeping the Bazel 
 ## Building libjsonnet.wasm
 
 ```bash
-GOOS=js GOARCH=wasm go build -o libjsonnet.wasm ./cmd/wasm 
+GOOS=js GOARCH=wasm go build -o libjsonnet.wasm ./cmd/wasm
 ```
 
 Or if using bazel:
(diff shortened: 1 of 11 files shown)


seh:
As the original author of this integration, I had been using this repository from a Bazel project that _didn't_ register a Go toolchain, as it was just trying to use the built _jsonnet_ tool, not integrate Jsonnet as a programming library (that is, as a dependency in my own Go code).

I understand this quandary, as I've run into myself in a different set of Bazel rules I've been writing. I think it warrants some coverage in the documentation to advise callers that they can't include this repository as a dependency without having a Go toolchain registered.

sparkprime:
So with this change we will be breaking people who have brought go-jsonnet into their bazel workspaces and who aren't already using Go?  I wonder how many people there are in that situation?

seh:
In my case, the motivation was to have _rules_jsonnet_ be able to rely on this Jsonnet port, as opposed to the C++ port. That was in #281, #290, and then #292.

psalaberria002:
#292 Added the call to `go_register_toolchains()`. That function seems to be safe to execute even when other go toolchains have been already defined. The call only fails when calling it with a specific version as in `go_register_toolchains(version="host")`. I am not a Bazel expert at all, so I am not sure what the best approach for configuring go toolchains is.

I have added the configuration steps to the Readme, and left the call to `go_register_toolchains()` in the dependencies, but I am not sure that belongs there. Any help is appreciated.

README.md
```diff
@@ -234,3 +234,37 @@ Similarly, after adding or removing Go source files, it may be necessary to sync
 ```bash
 bazel run //:gazelle
 ```
+
+## Using this repository from other Bazel workspaces
+
+To use `go-jsonnet` from a different Bazel workspace, add the repository
+and its dependencies to your WORKSPACE file.
+
+You will also need to register a Go toolchain.
```
sparkprime, line 243: Do you only need to do this if you don't already have one?

psalaberria002: That's right.

sparkprime:
Is this a problem with bazel versions maybe?

```
ERROR: /home/runner/.cache/bazel/_bazel_runner/d97bcaa4ec48514c66d33bdc5bb9b57b/external/google_jsonnet_go/bazel/deps.bzl:10:5: name 'go_register_toolchains' is not defined
ERROR: error loading package '': compilation of module 'bazel/deps.bzl' failed
```

sparkprime:
The log also has the line:

2022/05/24 19:16:12 Downloading https://releases.bazel.build/5.1.1/release/bazel-5.1.1-linux-x86_64...

psalaberria002:
Don't have time to work on this. Ended up building from source where I needed it.

PR #612, opened 2022-06-06
Update deps.bzl

AugustKarlstedt:
Fixes an issue reported in https://github.com/bazelbuild/rules_jsonnet/issues/125 where `com_github_fatih_color` does not specify its deps.

setting `build_external` to `external` allows bazel to call out to resolve imports. see https://github.com/bazelbuild/bazel-gazelle/blob/master/repository.md#go_repository


AugustKarlstedt:
there may be a better way to resolve this by using a different directive and specifying something like `build_directives` on the `com_github_fatih_color` `go_repository`

sbarzowski:
Hmmm... this build_external  thing looks a bit fishy to me, but I'm not super familiar with bazel. @sparkprime wdyt?

sparkprime:
I have no idea how any of this works

sparkprime:
Let's merge it and see if anyone complains

PR #613, opened 2022-06-13
handle int type in interpreter

rlankfo:
This PR adds support for `int` types in the interpreter. I ran into the need for this when attempting to evaluate a jsonnet file with input un-marshaled from YAML. 

YAML has support for `Integer` as a base type where JSON's numeric type is always a float.

### DIFF
--- a/interpreter.go
+++ b/interpreter.go
@@ -952,6 +952,8 @@ func jsonToValue(i *interpreter, v interface{}) (value, error) {
 		return makeValueBoolean(v), nil
 	case float64:
 		return makeDoubleCheck(i, v)
+	case int:
+		return makeDoubleCheck(i, float64(v))
 
 	case map[string]interface{}:
 		fieldMap := map[string]value{}
--- a/main_test.go
+++ b/main_test.go
@@ -96,6 +96,22 @@ var nativeError = &NativeFunction{
 	},
 }
 
+var jsonToValueRegr = &NativeFunction{
+	Name:   "jsonToValueRegr",
+	Params: ast.Identifiers{},
+	Func: func(_ []interface{}) (interface{}, error) {
+		m := make(map[string]interface{})
+		m["null"] = nil
+		m["array"] = []interface{}{"foo", "bar"}
+		m["boolean"] = true
+		m["float"] = 1.2
+		m["int"] = 1
+		m["object"] = map[string]interface{}{"key": "val"}
+		m["string"] = "string"
+		return m, nil
+	},
+}
+
 type jsonnetInput struct {
 	name             string
 	input            []byte
@@ -136,6 +152,7 @@ func runInternalJsonnet(i jsonnetInput) jsonnetResult {
 
 	vm.NativeFunction(jsonToString)
 	vm.NativeFunction(nativeError)
+	vm.NativeFunction(jsonToValueRegr)
 
 	rawAST, _, staticErr := parser.SnippetToRawAST(ast.DiagnosticFileName(i.name), "", string(i.input))
 	if staticErr != nil {
--- a/testdata/native_jsonToValueRegr.golden
+++ b/testdata/native_jsonToValueRegr.golden
@@ -1,0 +1,14 @@
+{
+  "array": [
+    "foo",
+    "bar"
+  ],
+  "boolean": true,
+  "float": 1.2,
+  "int": 1,
+  "null": null,
+  "object": {
+    "key": "val"
+  },
+  "string": "string"
+}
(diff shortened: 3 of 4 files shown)


sbarzowski:
Sorry, how exactly have you triggered this? Is there an example of JSON code (possibly with parseYaml) which hits this?

rlankfo:
@sbarzowski here's an extremely simplified example of what was happening:

main.go:
```go
// You can edit this code!
// Click here and start typing.
package main

import (
	"fmt"

	"github.com/google/go-jsonnet"
	"github.com/google/go-jsonnet/ast"
	"gopkg.in/yaml.v3"
)

var myYAML string = `
apiVersion: v1
data:
  username: YWRtaW4=
  password: MWYyZDFlMmU2N2Rm
  version: 42
kind: Secret
`

func main() {
	vm := jsonnet.MakeVM()

	vm.NativeFunction(&jsonnet.NativeFunction{
		Name:   "unmarshalYAML",
		Params: ast.Identifiers{"text"},
		Func:   unmarshalYAML,
	})
	vm.NativeFunction(&jsonnet.NativeFunction{
		Name:   "getSecret",
		Params: ast.Identifiers{"text"},
		Func:   getSecret,
	})
	res, err := vm.EvaluateFile("./sample.jsonnet")
	if err != nil {
		fmt.Println(err.Error())
		return
	}
	fmt.Println(res)	
}

func getSecret(_ []interface{}) (interface{}, error) {
	return myYAML, nil
}

func unmarshalYAML(i []interface{}) (interface{}, error) {
	text, ok := i[0].(string)
	if !ok {
		return nil, jsonnet.RuntimeError{Msg: "unmarshalYAML text argument must be a string"}
	}
	var v interface{}
	err := yaml.Unmarshal([]byte(text), &v)
	if err != nil {
		return nil, jsonnet.RuntimeError{Msg: err.Error()}
	}
	return v, nil
}

helpers.libsonnet:
```
{
  // fromYAML unmarshals YAML text into an object.
  fromYAML(text):: std.native('unmarshalYAML')(text),

  // example function that returns raw YAML
  getSecret(text):: std.native('getSecret')(text),
}
```

sample.jsonnet:
```
local helpers = import 'helpers.libsonnet';

{
  example: helpers.fromYAML(helpers.getSecret('example')),
}
```

If you run this code, you'll see it results in the following error:
```
RUNTIME ERROR: Not a json type: 42
        helpers.libsonnet:3:20-53       function <anonymous>
        sample.jsonnet:4:12-58  object <anonymous>
        Field "example"
        During manifestation
```

Removing the `version: 42` line from the "secret" YAML results in success.

**Comment by sbarzowski** [comment]
Why not use `parseYaml` that provided by Jsonnet which (presumably) doesn't have this problem instead of `unmarshalYAML`?

Generally, we expect `interface{}` result from native functions, so you can put anything there and we'll only complain at runtime (that's how Go deals with JSON data). Native Functions are supposed to return JSON, that's why we error out with `int` and it's expected.

That said, in this case we could extend it. I would add a regression test (might be something dumb like just returning an object with a field of each type). Here's the location where you would define it: https://github.com/google/go-jsonnet/blob/ece9c5dc949f90c5f6f5f84e34a32234b4b91c5e/main_test.go#L79.

**Comment by rlankfo** [comment]
@sbarzowski thanks for the feedback 🙏  I'll test out `parseYaml` to verify this works. I imagine it will work except in cases where we do custom unmarshalling (e.g. setting some default value if a field is not defined).  

For posterity, in case others run into this issue, I thought I'd mention that switching to the `sig.k8s.io/yaml` package instead of `gopkg.in/yaml.v3` may be preferred when the expected output is valid JSON because after converting to YAML it then uses `json.Marshal` and `json.Unmarshal` to convert to or from the struct.

In any case, I've included a regression test if you decide to include the change in this case. The test was failing when running locally and I'm not sure why -- the output indicated there was a diff with the golden file. It's not clear to me right off what the diff is:

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //:go_default_test
-----------------------------------------------------------------------------
--- FAIL: TestEval (0.26s)
    --- FAIL: TestEval/testdata/native_jsonToValueRegr (0.00s)
        main_test.go:371: golden file testdata/native_jsonToValueRegr.golden has diff:
            {
               "array": [
                  "foo",
                  "bar"
               ],
               "boolean": true,
               "float": 1.2,
               "int": 1,
               "null": null,
               "object": {
                  "key": "val"
               },
               "string": "string"
            }

FAIL
```

PR #614, opened 2022-06-30
feat: Report native function panics as runtime errors.

rudo-thomas:
Previously, you'd get an "INTERNAL ERROR: (CRASH)" with a full backtrace
and a link to file a bug against go-jsonnet.


sbarzowski:
I like this! Got a few bug reports this way which should be routed elsewhere.

PR #615, opened 2022-07-07
Add context.Context-aware variants of vm.Evaluate* methods, honor the context during evaluation and manifestation.

rudo-thomas:
The idea here is to be able to put some upper limit on the time it takes to process a jsonnet file (or any other piece of code). Previously, the best bet was to fork+exec with a timeout.

Please review this commit-by-commit:
 - the first one just mechanically moves arguments around (this was a TODO, I hope I got the idea right)
 - the second adds the methods and the context checks

The added test verifies that the manifestation respects the context, I don't have an idea about how to test the evaluation case. I'll be grateful for any hints :)

### DIFF
--- a/error_formatter.go
+++ b/error_formatter.go
@@ -18,6 +18,7 @@ package jsonnet
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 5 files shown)


rudo-thomas:
I've rebased this against the current master.

rudo-thomas:
Dropping the efforts to get this pulled, see the discussion at https://groups.google.com/g/jsonnet/c/c0rejf-Fo8Q/m/-PLNoewaAgAJ

PR #620, opened 2022-07-25
Don't pass version to `go_register_toolchains()` by default

### LINKED ISSUE #619: `jsonnet_go_dependencies()` fails if go toolchain has already been registered
In this commit https://github.com/google/go-jsonnet/commit/1b7cbff4cd267022109868c5fb1953f99e2f8355 `jsonnet_go_dependencies()` was modified to pass a version parameter to `go_register_toolchains()`. If a go toolchain has already been registered it results in this error:
```
Error in fail: go_register_toolchains: version set after go sdk rule declared (go_sdk)
```
And according to the docs, setting a version is disallowed if a toolchain has already been registered: https://github.com/bazelbuild/rules_go/blob/master/go/toolchains.rst#go-register-toolchains

glindstedt:
This should fix #619 where `jsonnet_go_dependencies()` fails because a go toolchain has already been registered.

### DIFF
--- a/bazel/deps.bzl
+++ b/bazel/deps.bzl
@@ -9,9 +9,12 @@ load(
     "go_repository",
 )
 
-def jsonnet_go_dependencies(go_sdk_version = "host"):
+def jsonnet_go_dependencies(go_sdk_version = None):
     go_rules_dependencies()
-    go_register_toolchains(version = go_sdk_version)
+    if go_sdk_version != None:
+        go_register_toolchains(version = go_sdk_version)
+    else:
+        go_register_toolchains()
     gazelle_dependencies()
     go_repository(
         name = "com_github_davecgh_go_spew",


sparkprime:
This gives an error: 
```
ERROR: Traceback (most recent call last):
	File "/home/runner/work/go-jsonnet/go-jsonnet/WORKSPACE", line 15, column 24, in <toplevel>
		jsonnet_go_dependencies()
	File "/home/runner/.cache/bazel/_bazel_runner/d97bcaa4ec48514c66d33bdc5bb9b57b/external/google_jsonnet_go/bazel/deps.bzl", line 17, column 31, in jsonnet_go_dependencies
		go_register_toolchains()
	File "/home/runner/.cache/bazel/_bazel_runner/d97bcaa4ec48514c66d33bdc5bb9b57b/external/io_bazel_rules_go/go/private/sdk.bzl", line 416, column 17, in go_register_toolchains
		fail('go_register_toolchains: version must be a string like "1.15.5" or "host"')
```

Issue #610, opened 2022-05-31
linter: std function unrecognized; skip lint directive?


bwplotka:
Hi, 

I think there is some linter problem with `std.manifestJsonMinified` function.

Code like:

```
local objects = [{ test: '' }, { bar: 'filter2' }, { bar: 'filter' }, { foo: '' }];

{
  spec: std.sort(objects, std.manifestJsonMinified),
}
```

Fails the lint with:

```
test.jsonnet:4:27-51 Indexed object has no field "manifestJsonMinified"

  spec: std.sort(objects, std.manifestJsonMinified),
```

Blocking: https://github.com/rhobs/configuration/pull/227

On a separate note, is there a way to add some directive that skips false positives?

sbarzowski:
> On a separate note, is there a way to add some directive that skips false positives? 

You mean like for a single line? That would be very useful, but we don't have that yet.

jaymebrd:
@bwplotka @sbarzowski I put up a PR to fix the linter issue

sbarzowski:
The PR was merged. Thanks a lot!

PR #621, opened 2022-07-28
feat: add support to the jsonnet linter for std.manifestJsonMinified

jaymebrd:
Adds support for the `std.manifestJsonMinified` to the linter

Helps #610


sbarzowski:
Thank you!

PR #624, opened 2022-09-02
feat: go impl parseint

jaymebrd:
This is a continuation of the effort to write the standard library in go-jsonnet in native go for performance improvement

This implements  `std.parseInt` into native Go instead of interpreted Jsonnet which results in a performance improvement of ~80 percent

```
➜  ./benchmark.sh Builtin                                                                                                                    
Running Before Test... (10s)
Running After Test... (10s)
benchmark                                  old ns/op      new ns/op      delta
Benchmark_Builtin_parseInt-12              30495144       6554907        -78.51%
```

Helps with of https://github.com/google/go-jsonnet/issues/111


sbarzowski:
Brilliant, thank you

PR #628, opened 2022-09-16
perf: add size hint to map creation

Jesse-Cameron:
Hey folks, just a small patch to add size hints to maps where it seemed appropriate. I'm not sure if I'm missing an obvious reason these were left out initially. But given `go-jsonnets` use of maps this could be a low hanging fruit for cutting down on allocations.


sbarzowski:
Looks good, thank you!

PR #629, opened 2022-09-20
perf: use native builtin's for foldl and foldr

Jesse-Cameron:
👋 

**Description**

When performing some profiling of one of our applications we found that calls to fold were quite slow. Using a native function is quite simple and gives some nice performance benefits.

**Benchmark results:**
```
benchmark                                 old ns/op      new ns/op      delta
Benchmark_Builtin_substr-8                17914126       16328579       -8.85%
Benchmark_Builtin_reverse-8               360776847      346691957      -3.90%
Benchmark_Builtin_parseInt-8              8686867        8314151        -4.29%
Benchmark_Builtin_base64Decode-8          22157223       21749268       -1.84%
Benchmark_Builtin_base64DecodeBytes-8     281975629      281841745      -0.05%
Benchmark_Builtin_base64-8                23824149       23716470       -0.45%
Benchmark_Builtin_base64_byte_array-8     141846327      141377054      -0.33%
Benchmark_Builtin_manifestJsonEx-8        9317781        9279067        -0.42%
Benchmark_Builtin_comparison-8            128783954      128362069      -0.33%
Benchmark_Builtin_comparison2-8           2082396660     2029123362     -2.56%
Benchmark_Builtin_foldl-8                 159908963      31278937       -80.44%

```

Note: I've left off a benchmark for `foldr` as I suspect it would be almost identical. Let me know if you want me to add one though.

**Issue**

https://github.com/google/go-jsonnet/issues/111


sbarzowski:
Good stuff!

Would you be able to add some more testcases? Our current ones are rather rudimentary and with a native implementation we would like various cases covered, in particular:
* Various types aggregated
* Empty and 1 element folding
* What if string is passed instead of an array?
It's enough to add example code as jsonnet file(s) in testdata and then run the following to generate golden outputs:
```
go test -update && go test ./linter -update
```
(Afterwards we need to check that the results are expected.)

Jesse-Cameron:
Hey @sbarzowski !

Good call on improving the coverage here. It actually uncovered an edge case I hadn't covered (when the second argument is a string). I've patched this and added the requested tests.

LMK if there are other areas you think could benefit from a test. TY 🙇‍♂️

sbarzowski:
All good, thank you!

PR #630, opened 2022-09-20
fix: remove go 1.13 → 1.16 as it's incompatible with sys

Jesse-Cameron:
according to maintainers, they only support the two latest versions:
https://github.com/golang/go/issues/55078#issuecomment-1247321657


sbarzowski:
Thank you!

PR #635, opened 2022-10-11
memory align structs

Jesse-Cameron:
There is a small amount of heap allocations that can be shaved off by making sure that all of the structs are aligned correctly. I've rearranged structs to ensure they are padded correctly and added a lint rule to enforce this in the future.

There are few ignored structs as I don't think contribute to the runtime memory usage. They could also be fixed up at a later time.

I've also fixed a small amount of other lint issues.

References: 
 - https://betterprogramming.pub/how-to-speed-up-your-struct-in-golang-76b846209587
 - https://go101.org/article/memory-layout.html

### DIFF
--- a/ast/fodder.go
+++ b/ast/fodder.go
@@ -66,10 +66,10 @@ const (
 
 // FodderElement is a single piece of fodder.
 type FodderElement struct {
+	Comment []string
 	Kind    FodderKind
 	Blanks  int
 	Indent  int
-	Comment []string
 }
 
 // MakeFodderElement is a helper function that checks some preconditions.
--- a/ast/location.go
+++ b/ast/location.go
@@ -187,13 +187,12 @@ func trimToLine(loc LocationRange, line int) LocationRange {
 
 // LineBeginning returns the part of a line directly before LocationRange
 // for example:
-// local x = foo()
 //
-//	^^^^^ <- LocationRange loc
-//
-// then
-// local x = foo()
-// ^^^^^^^^^^ <- lineBeginning(loc)
+//	local x = foo()
+//	          ^^^^^ <- LocationRange loc
+//	then
+//	local x = foo()
+//	^^^^^^^^^^ <- lineBeginning(loc)
 func LineBeginning(loc *LocationRange) LocationRange {
 	return LocationRange{
 		Begin:    Location{Line: loc.Begin.Line, Column: 1},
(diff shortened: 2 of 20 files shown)


ast/location.go
```diff
@@ -188,7 +188,9 @@ func trimToLine(loc LocationRange, line int) LocationRange {
 // LineBeginning returns the part of a line directly before LocationRange
 // for example:
 // local x = foo()
-//           ^^^^^ <- LocationRange loc
+//
+//	^^^^^ <- LocationRange loc
```
rudo-thomas, line 192: I have no approval powers here, but this comment got totally misformatted. Same many times below.

ast/fodder.go
```diff
@@ -66,10 +66,10 @@ const (
 
 // FodderElement is a single piece of fodder.
 type FodderElement struct {
+	Comment []string
```
sparkprime, line 69: Will this really affect performance? all 4 of these are 64 bits, not?

ast/location.go
Jesse-Cameron: Hey @rudo-thomas, I am using golang 1.19 in my environment. In 1.19 they updated how [godoc comments are formatted](https://tip.golang.org/doc/go1.19#go-doc). So I think that's where this change got pulled in from.

If we think that these formatting changes distract from the core of the PR, I'm happy to drop them. But anyone else that touches these files whilst using 1.19 is going to have similar issues.

ast/fodder.go
Jesse-Cameron: Thanks for looking over this @sparkprime, it might vary slightly bases on architecture. I think a slice of strings is larger than 8 bytes, on my machine it's 24. I'm guessing it's two ints for the `len` and `cap`. Then a pointer for the backing array.

To answer you're question: if it will affect performance. Perhaps not significantly. Running the benchmarks locally only reported slight improvements.

But given that this change also enabled [`fieldalignment`](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment) on this package, it was easy enough to go through an fix up the spots it reported the structs as misaligned.

ast/location.go
rudo-thomas: I'm saying these comments are now broken: The carets are no longer pointing where they used to – they don't make sense anymore.

The easiest would be to leave them unchanged. Possibly by using an earlier version of go fmt.

Jesse-Cameron: Ah yes. I see what you are saying. Thanks for explaining. I think it should be fairly easy to fix this using godoc codeblocks. I've updated here: 2b535a5

sbarzowski: @rudo-thomas For the record, I really appreciate when people spot issues in PRs!

sbarzowski:
Thank you! 

Technically, it breaks some public apis (AST stuff) and the bottleneck is about the number of objects allocated, not the size of structs, but I think it's still worth it.

Jesse-Cameron:
Thanks @sbarzowski !

> the bottleneck is about the number of objects allocated, not the size of structs

It certainly is! I was actually thinking of opening a GitHub issue to discuss this point :)

PR #637, opened 2022-10-17
feat: implementation of manifestTomlEx in Go (15 X faster)

jaymebrd:
Go implementation of `manifestTomlEx` - this was taking more than half the time in a large jsonnet repo. This implementation is ~15X faster

```
benchmark                                  old ns/op      new ns/op      delta
Benchmark_Builtin_manifestTomlEx-12        120747591      7949586        -93.42%
```

Helps with of https://github.com/google/go-jsonnet/issues/111


sbarzowski:
Good stuff, thanks!

PR #643, opened 2022-10-25
Update to latest CPP version

sparkprime:


anguslees:
See also #647.  Whatever order these are merged, we should ensure `bazel/repositories.bzl` is updated to match this new cpp_jsonnet git version.

PR #648, opened 2022-10-26
Add a CI test that exercises bazel builds

anguslees:
If it's not tested, then it doesn't work.  Add a CI test that
executes `bazel build //...` and `bazel test //...`, in addition to
the non-bazel build system.


anguslees:
Note bazel builds are currently broken.  We should merge this PR after the various other 'fix bazel' PRs.

sparkprime:
It seems the issue I'm having on my google issued machines is not happening in ci so it might be something I need to follow up with internally...  Then I will probably hit this `linter/internal/utils` issue.

sparkprime:
Oh I just noticed all the other PRs thanks so much for fixing all this stuff

Issue #656, opened 2022-11-22
std.member does not handle strings correctly


smcgivern:
Regular jsonnet defines `std.member` on strings in terms of `std.findSubstr`: https://github.com/google/jsonnet/blob/813c7412d1c7a42737724d011618d0fd7865bc69/stdlib/std.jsonnet#L236-L241

This means that if the first argument is a string, the second argument can be a string of any length.

The native version added in https://github.com/google/go-jsonnet/pull/638 only allows the second argument to be a single-character string. That is, with this input:

```jsonnet
{
  'foo': std.member('foobar', 'foo'),
}
```

`go-jsonnet` 0.18.0 returns:

```json
{
   "foo": true
}
```

But 0.19 and later returns:

```json
{
   "foo": false
}
```

Doing `std.member('foobar', 'o')` will return `true`, however.

salvacorts:
I just opened a PR fixing this: https://github.com/google/go-jsonnet/pull/657

smcgivern:
Thanks!

anguslees:
Request: can we get a 0.19.x, containing this fix?

It's subtle, is a regression, and yet silently does the wrong thing.  I just lost a few hours to debugging something that turned out to be this.

sparkprime:
Yes this makes sense and we should have done a release much earlier

PR #657, opened 2022-11-23
Fix std.member matching sub strings

salvacorts:
This PR fixes https://github.com/google/go-jsonnet/issues/656 and adds a test case.


sbarzowski:
Thank you!

Issue #634, opened 2022-10-07
Need readme instructions 'How to build' because I am not Go developer


CudaText-addons:
I need jsonnet-lint tool.
[jsonnet-lint](https://github.com/google/go-jsonnet/tree/master/linter).

I run 'go build' inside go-jsonnet dir. And what next? I got lot of files
```
total 484
drwxrwxr-x 18 user user  4096 окт  7 13:29 ./
drwxr-xr-x 81 user user 12288 окт  7 13:31 ../
drwxrwxr-x  2 user user  4096 окт  7 13:26 ast/
drwxrwxr-x  2 user user  4096 окт  7 13:26 astgen/
drwxrwxr-x  2 user user  4096 окт  7 13:26 bazel/
-rwxrwxr-x  1 user user   618 окт  7 13:26 benchmark.sh*
-rw-rw-r--  1 user user  1299 окт  7 13:26 BUILD.bazel
drwxrwxr-x  2 user user  4096 окт  7 13:26 builtin-benchmarks/
-rw-rw-r--  1 user user  1483 окт  7 13:26 builtins_benchmark_test.go
-rw-rw-r--  1 user user 49779 окт  7 13:26 builtins.go
drwxrwxr-x  2 user user  4096 окт  7 13:26 c-bindings/
drwxrwxr-x  3 user user  4096 окт  7 13:26 c-bindings-tests/
drwxrwxr-x  9 user user  4096 окт  7 13:26 cmd/
-rw-rw-r--  1 user user   266 окт  7 13:26 CONTRIBUTING
drwxrwxr-x  2 user user  4096 окт  7 13:26 cpp-jsonnet/
-rw-rw-r--  1 user user   979 окт  7 13:26 doc.go
-rw-rw-r--  1 user user  4347 окт  7 13:26 error_formatter.go
drwxrwxr-x  2 user user  4096 окт  7 13:26 formatter/
drwxrwxr-x  8 user user  4096 окт  7 13:29 .git/
drwxrwxr-x  3 user user  4096 окт  7 13:26 .github/
-rw-rw-r--  1 user user   432 окт  7 13:26 .gitignore
-rw-rw-r--  1 user user    91 окт  7 13:26 .gitmodules
-rw-rw-r--  1 user user   352 окт  7 13:26 .golangci.yml
-rw-rw-r--  1 user user   250 окт  7 13:26 go.mod
-rw-rw-r--  1 user user  1958 окт  7 13:26 .goreleaser.yml
[...]

PR #658, opened 2022-11-25
docs: add explicit mention for installing other tools in `cmd/`

jdockerty:
This would close https://github.com/google/go-jsonnet/issues/634 and avoid similar issues in the future.

It makes it clear than you can also install other tools, not only `jsonnet`, from the `cmd/` directory. I used `jsonnet-lint` as an example.


sbarzowski:
Thanks!

Issue #661, opened 2022-12-03
[1037] escapeStringXml was added to jsonnet; please make it available in go-jsonnet's std as well


scr-oath:
I added `std.stringEscapeXml` in github.com/google/jsonnet - is it maintained separately in this repo? How does one pull in and release updates to the stdlib in this library/executable?

See
- <https://github.com/google/jsonnet/pull/1037>
- <https://github.com/google/jsonnet/pull/1038>

PR #662, opened 2022-12-03
[fix 661] Update stdlib to get std.escapeStringXml

scr-oath:
See
- https://github.com/google/jsonnet/issues/1037
- https://github.com/google/jsonnet/pull/1038
- #661


sbarzowski:
Thank you!

PR #663, opened 2022-12-20
fix: Make lstripChars and rStripChars work with multibyte characters

iainlane:
We're currently working with the length of the string itself, which is the number of bytes. That's wrong in these functions which operate at the level of characters. When applied to a string with a multibyte character they can return the wrong result. For example, evaluating `std.rstripChars('• foo\n', '\n')` results in `"• f"`.
    
What we need to do instead is get the number of runes, and remove one rune at a time.

Fixes `std.lstripChars` and `std.rstripChars` when the source string contains multibyte (e.g. Unicode) characters.


iainlane:
I wanted to show that b44d7f765f9339037146a97a496e7c1f10519adb was failing and e08186a8ba2fc9d8fd4d24efea64ff131b3a055b passing but we need a manual approval before the checks will run, so I pushed both commits. :-)

iainlane:
@sbarzowski could you take a look at this please?

sbarzowski:
Linter test files need to be generated, otherwise looks good. You can just run `go test ./linter/ -update`.

(BTW these functions could be waaay faster if instead of calling the function recursively it found the cutoff index and then built the final string once).

iainlane:
> Linter test files need to be generated, otherwise looks good. You can just run go test ./linter/ -update. Thank you so much!

I've done it, cheers. They're empty, looks like that's expected based on the other files in there.

> (BTW these these functions could be waaay faster if instead of calling themselves recursively, they found the cutoff index and then built the final string once).

Yeah, I thought that too. It would be cool if they could just be reduced to `strings.Trim{Left,Right}` but it's a bit more complex because `chars` can be an array itself (not just a string).

PR #664, opened 2023-01-14
Adds jsonnet-lint to the releases

jregehr:
This PR should add jsonnet-lint to all the releases.

I haven't worked with a go releaser before. I think I figured out how the releases work for jsonnet and jsonnetfmt; I added similar stanzas for jsonnet-lint.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -42,6 +42,26 @@ builds:
     main: ./cmd/jsonnetfmt
     binary: jsonnetfmt
 
+  - env:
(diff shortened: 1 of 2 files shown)


sbarzowski:
This looks good to me. 

> I haven't worked with a go releaser before. I think I figured out how the releases work for jsonnet and jsonnetfmt; I added similar stanzas for jsonnet-lint.

I'm not aware of a better workflow :-). Let's see if it works.

PR #669, opened 2023-02-15
feat: implement std.sum

rohitjangid:
Add `std.sum` function in standard library.

PR for cpp-jsonnet: https://github.com/google/jsonnet/pull/1048


builtins.go
```diff
@@ -1471,7 +1471,7 @@ func tomlEncodeString(s string) string {
 }
 
 // tomlEncodeKey encodes a key - returning same string if it does not need quoting,
-// otherwise return it quoted; returns empty key as ''
+// otherwise return it quoted; returns empty key as ”
```
sparkprime, line 1474: Looks like your editor changed `''` to `"`

rohitjangid: Yes. Reverted this.

sparkprime:
I think this is good.  Can you also go to the cpp jsonnet repo where the stdlib and docs are, and do the following 3 small things:

1) Add a pure jsonnet version to `stdlib/std.jsonnet`, which can just look like this: ./jsonnet -e 'std.foldl(function(a,b)a+b,[1,2,3],0)'

2) Add the same test you put in this PR to the `test_suite/stdlib.jsonnet` file

3) Update the docs which are in `doc/_stdlib_gen/stdlib-content.jsonnet`

rohitjangid:
Created https://github.com/google/jsonnet/pull/1048 with above mentioned change

PR #683, opened 2023-04-11
feat: implement std.round

rohitjangid:
Add `std.round` function in standard library.

PR for jsonnet: https://github.com/google/jsonnet/pull/1064


rohitjangid:
@sparkprime Can you help me understand why these checks are failing? By looking at error messages, it looks like it something is failing in go sdk.

sparkprime:
there's something going on but also i think there's a trailing newline diff with your golden files.  `go test --update` will fix that.  However I'm just going to merge it and sort it all out myself because of the other issue.

PR #684, opened 2023-04-14
Implement std.trim for string

Tejesh-Raut:
Add `std.trim` function in standard library.

PR for cpp-jsonnet: https://github.com/google/jsonnet/pull/1067


Tejesh-Raut:
Tests added in this PR are working fine. Not sure why 2 other checks are failing. Can someone help here?

builtins.go
```diff
@@ -1289,6 +1289,15 @@ func builtinIsEmpty(i *interpreter, strv value) (value, error) {
 	return makeValueBoolean(len(sStr) == 0), nil
 }
 
+func builtinTrim(i *interpreter, strv value) (value, error) {
+	str, err := i.getString(strv)
+	if err != nil {
+		return nil, err
+	}
+	sStr := str.getGoString()
+	return makeValueString(strings.TrimSpace(sStr)), nil
```
sparkprime, line 1298: To be compatible with C++, this should be strings.Trim with explicit " \t" I think?

sparkprime:
I think the tests will work if you rebase.

builtins.go
Tejesh-Raut: `strings.TrimSpace` removes all kinds of spaces '\t', '\n', '\v', '\f', '\r', ' ', U+0085 (NEL), U+00A0 (NBSP).

PR #693, opened 2023-05-03
Resolves a false-positive detection of multi-doc YAML streams

### LINKED ISSUE #673: false-positive in YAML stream detection
It seems like that go-jsonnet detects a false-positive YAML stream when `std.parseYaml` parses a YAML document that contains `---` in data. It is also an inconsistent behavior with the C++ version.

# How to reproduce
Suppose that we have the next `test.jsonnet`:

```jsonnet
std.parseYaml('foo---: bar')
```

This is not a YAML stream, and it is not far from the example shown in [the std library reference](https://jsonnet.org/ref/stdlib.html#parseYaml). So I expect this input produces the next result:

```json
{
   "foo---": "bar"
}
```

The C++ implementation actually produces this expected result. However, go-jsonnet produces the next output.

```console
$ jsonnet -v
Jsonnet commandline interpreter (Go implementation) v0.19.1
$ jsonnet test.jsonnet
[
   {
      "foo---": "bar"
   }
]
```

I believe this is because `builtinParseYAML` ([builtins.go#L1375](https://github.com/google/go-jsonnet/blob/83b9f8341f65a054b046bcba5a3092227cddccf9/builtins.go#L1375)) detects YAML streams too loosely.

yugui:
Fixes #673.


sparkprime:
Nice

PR #695, opened 2023-05-05
Adds std.any and std.all to jsonnet-lint

jregehr:
I did not add test cases. I haven't learned how the testing language works yet. Please build and test.

Fix for Issue  #1080


sparkprime:
I don't think we have tests for each specific std function and I think it'd be a bit toily to do that.  They really ought to be taken from a declarative config of some sort, or perhaps annotations in the code.  So I'll just accept this as is.

PR #698, opened 2023-05-10
Add support for bzlmod

mortenmj:
This PR adds a MODULE.bazel file which allows other repositories to import this one as a bzlmod module.
I don't know how Google handles publishing to the Bazel Central Registry, so I haven't made any attempt at that.

This is ready for review, but not for merging, pending https://github.com/google/jsonnet/pull/1083 which adds bzlmod support for jsonnet. Jsonnet would need to be added to the Bazel Central Registry for this PR to be merged, as bzlmod does not allow importing modules that use overrides (such as `git_override` used here).

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,20 @@
+module(name = "jsonnet_go", version = "0.0.0")
+
+bazel_dep(name = "gazelle", version = "0.30.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "jsonnet", version = "0.0.0", repo_name = "cpp_jsonnet")
+bazel_dep(name = "rules_go", version = "0.39.1", repo_name = "io_bazel_rules_go")
+
+git_override(
+    module_name = "jsonnet",
+    remote = "https://github.com/mortenmj/jsonnet.git",
+    commit = "df811822d9a31cb3fab9bdf68679ad300f1c12ed",
+)
+
+go_deps = use_extension("@bazel_gazelle//:extensions.bzl", "go_deps")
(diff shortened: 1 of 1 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,20 @@
...
+
+bazel_dep(name = "gazelle", version = "0.30.0", repo_name = "bazel_gazelle")
+bazel_dep(name = "jsonnet", version = "0.0.0", repo_name = "cpp_jsonnet")
+bazel_dep(name = "rules_go", version = "0.39.1", repo_name = "io_bazel_rules_go")
+
+git_override(
+    module_name = "jsonnet",
+    remote = "https://github.com/mortenmj/jsonnet.git",
+    commit = "df811822d9a31cb3fab9bdf68679ad300f1c12ed",
+)
```
mortenmj, line 11: This would be updated once https://github.com/google/jsonnet/pull/1083 lands. This PR is not ready until that happens.

sparkprime:
LGTM

sparkprime:
(Either for people using this or maintaining it)

mortenmj:
@sparkprime I am not aware of any documentation that goes into how to import jsonnet/go-jsonnet into a WORKSPACE, but I'd want to add similar documentation for using MODULE.bazel.

I think it could be a good idea to add a test that shows how this works. This could then also be used in the BCR presubmit.yml, which would add some confidence that we don't break things in the future. Here's an example of what that might look like:
https://github.com/protocolbuffers/protobuf/tree/main/examples
https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/3.19.6/presubmit.yml

PR #701, opened 2023-05-25
feat: Add std.parseCav and std.jsonToCsv

rohitjangid:
Implement `std.parseCsv` and `std.jsonToCsv` in standard library


builtins.go
```diff
@@ -1425,6 +1426,143 @@ func builtinParseYAML(i *interpreter, str value) (value, error) {
...
+	var keys []string
+
+	reader := csv.NewReader(strings.NewReader(s))
+	for row := 0; ; row++ {
+		record, err := reader.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, i.Error(fmt.Sprintf("failed to parse JSON: %s", err.Error()))
```
sparkprime, line 1446: s/JSON/CSV ?

builtins.go
```diff
@@ -1425,6 +1426,143 @@ func builtinParseYAML(i *interpreter, str value) (value, error) {
...
+		}
+
+		if row == 0 { // consider first row as header
+			keys = record
+		} else {
+			j := make(map[string]interface{})
+			for i, k := range keys {
+				j[k] = record[i]
```
sparkprime, line 1454: This seems a bit opinionated.  Is there a precedent for this way of doing it?  E.g. what if the column header is the empty string or just that 2 columns have a duplicate column header.  Seems like this would only let you parse a CSV it it followed a pattern of having unique column headers.

sparkprime: If you want that semantics it'd be better to call it `parseCsvWithHeader` or something like that.  Then we could add a more generic one that just returns an array of arrays as `parseCsv` later.

sparkprime: another option: `parseCsvToObjects`

builtins.go
```diff
@@ -1425,6 +1426,170 @@ func builtinParseYAML(i *interpreter, str value) (value, error) {
...
+			return nil, i.Error(fmt.Sprintf("failed to parse CSV: %s", err.Error()))
+		}
+
+		if row == 0 { // consider first row as header
+			// detect and handle duplicate headers
+			keyCount := map[string]int{}
+			for _, k := range record {
+				keyCount[k]++
+				if c := keyCount[k]; c > 1 {
+					keys = append(keys, fmt.Sprintf("%s__%d", k, c-1))
```
sparkprime, line 1473: I think this is actually going to be confusing because it's not going to round-trip back through the manifest function (the mangled column names will appear in the output).  Also the behaviour would have to be documented which would be a pain.  Now you've renamed the function and it doesn't look like a generic CSV reader anymore I think what you were originally doing was fine.

rohitjangid: I guess we can add an argument called `handle_duplicate_headers=false` for users to opt for this. Using this flag, we could either overwrite the duplicate headers or rename the duplicate fields as above. WDYT?

sparkprime: SGTM

rohitjangid: Made the changes

rohitjangid:
Hi @sparkprime 
Can we merge this if there are no other concerns?

rohitjangid:
@sparkprime nudging this PR

PR #704, opened 2023-05-31
feat: Implement std.parseXmlJsonml

rohitjangid:
Implement `std.parseXmlJsonml` as standard function


jsonml.go
```diff
@@ -0,0 +1,145 @@
...
+			for _, a := range t.Attr {
+				attr[a.Name.Local] = a.Value
+			}
+			node = append(node, attr)
+		}
+		b.stack.Push(node)
+		b.currDepth++
+	case xml.CharData:
+		t := token.(xml.CharData)
+		s := strings.TrimSpace(string(t))
```
sparkprime, line 111: Is there a reason for trimming it?  I think whitespace is actually meaningful in XML because of things like `<b>foo</b>bar` vs `<b>foo</b> bar`.  If we're only supporting the (admittedly common) case of the XML file being whitespace agnostic (i.e. whoever consumes the XML doesn't care about the whitespace) then we can document it as not supporting that.  However I think just not trimming the strings here will fix it, will it not?

rohitjangid: The reason behind trimming whitespaces is to simplify jsonml. Currently due to formatted xml, jsonml becomes bit complex to work with due to empty content. We want to ignore empty lines but I agree current implementation would also trim spaces around the content. I could take of it.

I am thinking of allowing users to control this behaviour. We can provide an argument `trimWhitespaces = true` to the function which would control if user wants to remove formatting whitespaces. 

WDYT?

sparkprime: I found this which looks interesting.  http://xml.coverpages.org/rfc-wshp19990416.html  I wonder if the current behaviour of the manifester is wrong?

rohitjangid: What I understand from these link is it is safe to remove boundary whitespaces as a default behaviour. We would provide an argument `preserveWhitespaces = false` to preserve them if anyone wants to do so.

sparkprime: There seems to be some diversity in what other parsers are doing so I think anything we do is fine as long as it's documented.

rohitjangid: Updated

rohitjangid:
Hi @sparkprime 
Can we merge this if there are no other concerns?

rohitjangid:
@sparkprime nudging this PR

PR #705, opened 2023-06-02
Teach jsonnet-lint about optional parameters of std.manifestYamlDoc

swgillespie:
manifestYamlDoc takes two optional parameters, `indent_array_in_object` and `quote_keys`. This commit teaches jsonnet-lint about them so that it doesn't raise errors when you use them.

There are other stdlib library functions with this problem; the true solution is probably to auto-generate this from the stdlib AST, but this at least gets the linter happy with this particular function.


sparkprime:
There is also the documentation which exists as a jsonnet file and I believe lists the params.

PR #530, opened 2021-05-06
Expose "unparse", FormatNode and RawAST

tomwilkie:
(Opening this for discussion, see #389) 

I've build a little tool that does some modifications to jsonnet files (updating docker image versions in them), and wanted to use the pretty-printing code to write out the modified AST.  

This is a minimal set of changes that allow me to do that.

Issue #389, opened 2020-03-07
Figure out what code-modifying tools should do about formatting


sbarzowski:
Currently there are no code-modifying tools, other than the autoformatter that I know about, but we may want to add autofix to linter one day. Some 3rd party users also expressed interest in writing code-manipulating tools.

This is finally possible with the addition of Fodder to the AST. Unfortunately it is still not clear how to actually save the AST back in the text form. Using autoformatter to dump AST requires being aware of the formatting options. Also, using a pinned version of the formatter makes sense. 

Perhaps we can have a near-full-fidelity, optionless unparser? 
The tools could then change the affected part of the code, without affecting the rest too much. Implementing such unparser may require changing the definition of the Fodder and saving some additional info in the AST.

sparkprime:
Currently I think we force indentation by spaces, strip whitespace at the end of lines, and compress whitespace between tokens on a line to a single char (except for before the comment on the end of a line, which is always 2 spaces). I think that covers it?  Aware of any more?

sbarzowski:
There is also configurable array and object padding: https://github.com/google/go-jsonnet/blob/master/internal/formatter/unparser.go#L288.

sparkprime:
I wonder if we should get rid of all these options (or at least not expose them on the commandline).  The other option is to infer it from the input, but then what to do if the file has a mix of them or something weird like `[1 ]`.

Do people really use the formatter options?

On the other hand, it would be a regression vs C++ not to support these options.

sh0rez:
I'm not sure if people really use them (we set some, not sure why), but I think what matters more is that we now have the chance to do some breaking changes here, as we basically release a formatter (sure, there was a c++ version, but it had no widespread use).

That being said, I'd like to see something similar to `gofmt`:
**Nobody loves `gofmt`'s style, yet everybody loves gofmt**. This is important here. After all it doesn't matter how the style looks, as long as it is not super unpopular, it's most important that it is consistent. That we can expect every single `.jsonnet` file to be formatted according to this.

As an example, I don't really care if we had paddings (I prefer not to have them, but I don't care after all), but I do care that everybody does the same. Without thinking about it.

sparkprime:
Cpp formatter did get lots of use for people using Jsonnet before go-jsonnet was a better alternative.  I'm curious to know how many people since missed out on autoformatting because they didn't know about or didn't want to build the C++ version...

heuristicAL:
[...]

PR #710, opened 2023-06-14
Expose "unparse", FormatNode and RawAST

charleskorn:
This is a rebased version of #530 by @tomwilkie:

> (Opening this for discussion, see https://github.com/google/go-jsonnet/issues/389)
> 
> I've build a little tool that does some modifications to jsonnet files (updating docker image versions in them), and wanted to use the pretty-printing code to write out the modified AST.
> 
> This is a minimal set of changes that allow me to do that.

### DIFF
--- a/formatter/formatter.go
+++ b/formatter/formatter.go
@@ -4,7 +4,10 @@
 // customized using formatter.Options.
 package formatter
 
-import "github.com/google/go-jsonnet/internal/formatter"
+import (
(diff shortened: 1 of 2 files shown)


sparkprime:
This `does not produce any go_sdk headers files` failure I've seen before.  I think I fixed it by bumping the version of the bazel/golang integration package that we use.

PR #715, opened 2023-06-27
feat: implement std.match as standard function

rohitjangid:
Implement `std.match` in standard library

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -29,6 +29,7 @@ import (
 	"io"
 	"math"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


sparkprime:
I was hoping to merge https://github.com/google/jsonnet/pull/1039 this release, but I don't think we have a go-jsonnet equivalent yet (@sbarzowski is that right?)  Apparently the stdlib regexes in Go are similar enough that we can consider them compatible, in which case maybe all you have to do is make sure the API has the same function name / param name as the existing PR?

rohitjangid:
If that is the case, I could create a separate PR with equivalent implementation of all the api in go. Let me know if I should go ahead with that considering cpp PR is yet to be merged.

rohitjangid:
I didn't found any documentation for https://github.com/google/jsonnet/pull/1039
Is that part of any other PR?

sbarzowski:
Very cool that this is happening!

Yeah, both this and https://github.com/google/jsonnet/pull/1039 are RE2 based, so I'd just use go standard library here. There are minor differences, but full regexp compatibility beyond simple cases cannot be guaranteed anyway – they are not specified precisely enough.

> I didn't found any documentation for https://github.com/google/jsonnet/pull/1039

it doesn't have docs (yet), but you can see the summary in the PR description. If we could have the functions as described there, it would be great.

PR #730, opened 2023-09-24
Change FindDependencies to optionally follow symlinks

thequailman:
I am using the FindDependencies function in this library to gather jsonnet files for future rendering.  The current implementation of getAbsPath follows symlinks, which makes it very hard to determine later on what the file path should be when using a custom Importer.  This change adds a flag to getAbsPath to make the symlink resolution optional.  jsonnet-deps will still follow symlinks/there should be no behavior change for users.

### DIFF
--- a/cmd/jsonnet-deps/cmd.go
+++ b/cmd/jsonnet-deps/cmd.go
@@ -193,7 +193,7 @@ func main() {
 		}
 	}
 
-	dependencies, err := vm.FindDependencies("", conf.inputFiles)
+	dependencies, err := vm.FindDependencies("", conf.inputFiles, true)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
--- a/vm.go
+++ b/vm.go
@@ -232,8 +232,11 @@ func (vm *VM) evaluateSnippet(diagnosticFileName ast.DiagnosticFileName, filenam
 	return output, nil
 }
 
-func getAbsPath(path string) (string, error) {
+func getAbsPath(path string, followSymlinks bool) (string, error) {
 	var absPath string
+
+	var err error
+
 	if filepath.IsAbs(path) {
 		absPath = path
 	} else {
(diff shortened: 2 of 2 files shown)


thequailman:
This should be ready to merge, if possible.  We're currently maintaining a fork with this functionality and would love to remove the `replace` line in our `go.mod`.

thequailman:
Hello:

> Is this still a use-case that you need? Are you still maintaining your own fork for this?

Yes, we still need this functionality.  All of our products use Jsonnet, but one in particular, [Etcha](https://etcha.dev), builds/packages artifacts from various Jsonnet files.  Avoiding the symlink resolution ensures we can properly walk various files and generate a tree of files to be packaged/imported.  If we have to use symlink resolution, we end up importing/bundling the same file multiple times.  We could work around this with hashing or dedupe algorithm, but just avoiding symlink resolution is far easier for us.

> So this suggests you're using the FileImporter when looking for dependencies, and then using a different importer when you actually evaluate the inputs - is that right?

That's right, we basically do FindDependencies, gather the files, determine entry points, and store them in a `map[<relative file path>]contents` for bundling.

> As an alternative to removing the EvalSymlinks call we could keep the "original" untranslated path in the deps map (so it is keyed by canonical path but holds the original path), and then we can report the untranslated path while retaining the same deduplication behaviour.

This could work maybe, but I'm not sure how we'd get this from FindDependencies.  It returns a []string with the filepaths.

thequailman:
Exactly, appreciate you taking the time to understand the use case here.  I'm fine using a different function from this library, it's a couple line change in our internal jsonnet library that abstracts this one.  I could even change this PR to add FindDependenciesRelative instead that just doesn't follow symlinks so it doesn't break the API.  Whatever works for you folks, I know my use case is a little niche here so appreciate however it ends up being included.

johnbartholomew:
Looks good. Could you run `go fmt ./vm.go` and amend the last commit?

PR #739, opened 2023-12-28
feat: add debugger support

theSuess:
This PR adds an extension point for future debugger implementations.

We first investigated having the entire debugger codebase only use go-jsonnet as a dependency, but many of the fields and types needed to implement this are not exposed to external packages.

To solve this, we added a new `Debugger` type, which allows debugging frontends to control the evaluation of a specific file.

The API type is heavily inspired by the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/) and can be used to implement basic debugging functionality like:

* Stepping through evaluation
* Setting Breakpoints
* Automatically breaking on exceptions
* Viewing a rudimentary stack trace
* Inspecting variables at runtime

Example implementations of clients using this interface can be found in the [example-clients branch of our fork](https://github.com/grafana/go-jsonnet-debugger/tree/feat/example-clients/cmd/jsonnice).

The following recording shows the basic capabilities in action by using the DAP example client with VSCode:

https://github.com/google/go-jsonnet/assets/1364576/f2f6c2a4-32c1-468b-bc93-1d9c18f2a665

### DIFF
--- a/debugger.go
+++ b/debugger.go
@@ -1,0 +1,401 @@
+package jsonnet
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/go-jsonnet/ast"
+	"github.com/google/go-jsonnet/toolutils"
+)
+
+type Debugger struct {
+	// VM evaluating the input
+	vm *VM
+
+	// Interpreter built by the evaluation. Required to look up variables and stack traces
+	interpreter *interpreter
+
+	// breakpoints are stored as the result of the .String function of
+	// *ast.LocationRange to speed up lookup
+	breakpoints map[string]bool
+
+	// The events channel is used to communicate events happening in the VM with the debugger
+	events chan DebugEvent
+	// The cont channel is used to pass continuation events from the frontend to the VM
+	cont chan continuationEvent
+
+	// lastEvaluation stores the result of the last evaluated node
+	lastEvaluation value
+
+	// breakOnNode allows the debugger to request continuation until after a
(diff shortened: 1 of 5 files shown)


theSuess:
We have released an implementation of the example client as a standalone package at [grafana/jsonnet-debugger](https://github.com/grafana/jsonnet-debugger).

It is also integrated into the latest version of our [vscode-jsonnet](https://github.com/grafana/vscode-jsonnet) extension.

The debugger uses the [grafana/go-jsonnet-debugger](https://github.com/grafana/go-jsonnet-debugger) fork to implement the functionality until the PR is merged

interpreter.go
```diff
@@ -287,6 +308,13 @@ func (i *interpreter) newCall(env environment, trimmable bool) error {
 }
 
 func (i *interpreter) evaluate(a ast.Node, tc tailCallStatus) (value, error) {
+	i.evalHook.pre(i, a)
```
sbarzowski, line 311: I would probably make evalHook a pointer and avoid a virtual call here if it's null (so that the performance overhead when not debugging is smaller). Not sure how significant that is, given other overhead.

debugger.go
```diff
@@ -0,0 +1,401 @@
+package jsonnet
```
sbarzowski, line 1: There's quite a bit of complexity here - some tests would be nice.

sbarzowski (approved):

sbarzowski:
Really cool stuff.

Added some comments for potential follow-up, but merging as it is.

PR #741, opened 2024-01-11
Fix interface conversion from int types to float64

shrdlu68:


suzuki-shunsuke:
Can we get a chance to review this pull request?
I wrote an issue about this pull request.

- https://github.com/google/go-jsonnet/issues/752

interpreter.go
```diff
@@ -950,7 +950,8 @@ func jsonToValue(i *interpreter, v interface{}) (value, error) {
 	case bool:
 		return makeValueBoolean(v), nil
 	case int, int8, int16, int32, int64:
-		return makeDoubleCheck(i, v.(float64))
+		val, _ := v.(int64)
```
suzuki-shunsuke, line 953: I'm not sure why we need to convert the value to `int64`.

Is the following code wrong?

```go
return makeDoubleCheck(i, float64(v))
```

suzuki-shunsuke:
I'm trying this pull request, but this doesn't work as expected.
`panic` doesn't occur but the return value is always `0`.
I'm not sure why.

interpreter.go
suzuki-shunsuke: I see.

`makeDoubleCheck(i, float64(v))` causes a compile error.

```
cannot convert v (variable of type interface{}) to type float64: need type assertion
```

sbarzowski:
Closing in favour of https://github.com/google/go-jsonnet/pull/753

PR #742, opened 2024-01-26
feat: Go implementation for manifestYamlDoc and escapeStringJson

jgraeger:
This PR introduces native Go implementations for two standard library functions in Jsonnet: manifestYamlDoc and escapeStringJson. These enhancements are a response to performance challenges, especially noticeable in cases where `quote_keys=false` is set. This issue is documented in google/jsonnet#1019 and also holds for the Go implementation.

## Benchmarks
```
Running Before Test... (10s)
Running After Test... (10s)
benchmark                                old ns/op     new ns/op     delta
Benchmark_Builtin_escapeStringJson-8     9061562       3082901       -65.98%
Benchmark_Builtin_manifestYamlDoc-8      12660395      3559494       -71.88%
```

On a large, non-public codebase generating megabytes of yaml from jsonnet, the runtime decreased from over 10 minutes to 8 seconds.


builtins.go
```diff
@@ -1980,6 +2002,209 @@ func builtinManifestJSONEx(i *interpreter, arguments []value) (value, error) {
...
+						buf.WriteString(cindent)
+						buf.WriteString(yamlIndent)
+						cindent = cindent + yamlIndent
+					} else {
+						buf.WriteByte(' ')
+					}
+				} else {
+					buf.WriteByte(' ')
+				}
+				aux(fieldValue, buf, cindent)
```
rudo-thomas, line 2194: Note that this is unchecked.

Fix: https://github.com/google/go-jsonnet/pull/800

rudo-thomas: This PR (#742) introduces a bug into manifestYamlDoc, which is now part of the 0.21.0 release. (We were even getting stack traces, but that's besides the point here.)

The fix is trivial (#800), but I'm having difficulties getting it merged -- nobody responded...

@jgraeger, can I kindly ask you, as the original author, to help me ping the right people and get the fix merged? Thanks :)

PR #746, opened 2024-02-24
Remove dependency on gen to generate sets

itchyny:
Previously back in 2017, the set implementations stopped updating due to https://github.com/google/go-jsonnet/pull/95. This PR looks a bit thoughtless, and we should decide how to deal with the code generated by gen. As for now, https://github.com/clipperhouse/gen has been deprecated, and also https://github.com/clipperhouse/set has been renamed to `setwriter`. Right now the command `make install.dependencies` does not work in Go 1.22, because `go get` no longer supports legacy `GOPATH` mode (see https://tip.golang.org/doc/go1.22#go-command). The set implementations are straightforward, so I would like to suggest to stop using the `gen` tool to generate set types.


sbarzowski:
Yeah, we should avoid using codegen like this. 

(We still want codegen for stdlib AST - to avoid parsing overhead on startup, but that's maintained by us and less hairy).

PR #756, opened 2024-06-10
chore: update rules_go and gazelle version to fix bazel build

itchyny:
This PR should fix the `bazel test` job on CI.


itchyny:
My bad, #746 was missing updates in `ast/BUILD.bazel`.

itchyny:
Yey, `bazel test` is back to green.

PR #759, opened 2024-06-11
fix: use debugValueToString for debugger operations

theSuess:
The `valueToString` operation introduced by #742 is incompatible with the way the implementation from #739 as it tries to manifest an object from stack while the implementation needed by the debugger returns the value as-is without further evaluation.


sbarzowski:
Thank you!

PR #770, opened 2024-09-30
linter: Add std.clamp

nate-thirdwave:
Added missing linter definition for std.clamp


johnbartholomew:
I'm sorry for the lack of response on this PR. This has been updated by #840 which I've just merged, so this is now done.

PR #772, opened 2024-10-08
Add SetVM and GetVM methods to Debugger for TLA and extCode integration

vasusheoran:
This PR introduces two new methods, `SetVM` and `GetVM`, to the `Debugger` struct in `debugger.go`. These methods provide access to the underlying `VM` instance associated with the debugger:

* **`SetVM(vm *VM)`:** Sets the `VM` instance to be used by the debugger. This allows for dynamic configuration or replacement of the VM during the debugging process, enabling the integration of TLA and ext functionality.
* **`GetVM() *VM`:** Returns the current `VM` instance being used by the debugger. This can be useful for inspecting or modifying the VM's state.

The addition of these methods enhances the flexibility and control of the go-jsonnet debugger, making it easier to integrate TLA and ext features.

### DIFF
--- a/debugger.go
+++ b/debugger.go
@@ -345,6 +345,16 @@ func (d *Debugger) StackTrace() []TraceFrame {
 	return trace
 }
 
+func (d *Debugger) SetVM(vm *VM) {
+	if vm != nil {
+		d.vm = vm
+	}
+}
+
+func (d *Debugger) GetVM() *VM {
+	return d.vm
+}
+
 func debugValueToString(v value) string {
 	switch i := v.(type) {
 	case *valueFlatString:


vasusheoran:
@sbarzowski Can you please take a look?

lorenzboguhn:
Hi @johnbartholomew, 
I really would like to use the jsonnet debugger with external variables. 
This could be an approach to add them to the vm. 
If this approach here is not considered to be appropriate for the debuger, I would like to know if there is any interests on a PR that would add `extVars` or `extCode` similar to how external Libraries(jPaths) are added. 
This could be immensely helpful and I would try to implement it.
I mentioned you because you merged something 5 days ago.


Thanks for reading, best regards,
Lorenz

puremourning:
I did a quick change for my own personal use. I think it's a breaking API change, so probably needs a small refactor, but here it is:

* https://github.com/puremourning/go-jsonnet/commit/7cbb8db485d855835a639aba34e89478eb19af64
* https://github.com/puremourning/jsonnet-debugger/commit/22628345a0b3508f1d2f994a7290cb4d160208fb

PR #777, opened 2024-10-23
Add support for negative array indexes

ricardbejarano:


ricardbejarano:
Anyone? I think this is pretty straightforward to evaluate.

PR #782, opened 2025-01-18
Update stdlib and add builtins for atan2, hypot

johnbartholomew:
This brings stdlib up to the latest (unreleased) version, which includes some new functions added in https://github.com/google/jsonnet/pull/1187

Also adds builtins for `atan2` and `hypot` to match the corresponding builtins in the C++ version.

The full stdlib changes can be seen from the changes to std.jsonnet at: https://github.com/google/jsonnet/compare/f45e01d632b29e4c0757ec7ba188ce759298e6d3...4487bfb8ba7ed7a3d91d79d4a1aa2205e7839558

(That shows all changes to the C++ repo from the previously embedded version to the newly embedded version)

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -4,8 +4,8 @@ load(
 )
 
 # NB: update_cpp_jsonnet.sh looks for these.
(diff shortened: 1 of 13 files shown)


johnbartholomew:
Some checks are failing but that looks like unrelated breakage of the bazel build.

Issue #790, opened 2025-02-28
`std.maxArray` and `std.minArray` do not produce the correct result when passed a non-identity `keyF` function


eduardosm:
For example,
```jsonnet
std.maxArray(["a1", "b2"], keyF=function(x) x[1])
```
outputs `"2"`, when it should output `"b2"`.

The C++ version works correctly.

johnbartholomew:
Wow, good catch. Thanks!

PR #791, opened 2025-02-28
fix: output original value from array in `std.minArray` and `std.maxArray` instead of result of `keyF`

eduardosm:
Fixes https://github.com/google/go-jsonnet/issues/790.


johnbartholomew:
Thanks for finding and fixing this!

johnbartholomew:
Rebased, added support for the optional `onEmpty` parameter (for parity with the C++ version which already had this), and merged.

Thanks again for the contribution!

PR #792, opened 2025-03-06
Fix an out-of-bounds array access when formatting some empty comment

mickael-carl:
Also make sure to use default options in tests, to be consistent with what jsonnetfmt actually does.

Quick reproducer of the issue:
```bash
$ echo "#
  {}" | jsonnetfmt -
panic: runtime error: index out of range [1] with length 1
```


johnbartholomew:
Thanks for finding and fixing this!

PR #794, opened 2025-03-07
chore: Update dependency versions

andrii-korotkov-verkada:
Bump go and dependencies to the versions close to the latest.


johnbartholomew:
I think the Bazel build can use Go 1.23 by explicitly pulling in the right version using rules_go. In `MODULE.bazel` , after the `bazel_dep` for rules_go, add:

```
go_sdk = use_extension("@io_bazel_rules_go//go:extensions.bzl", "go_sdk")
go_sdk.download(
    name = "go_sdk",
    version = "1.23.7",
)
```

Probably ok to also bump the bazel module deps too, rules_go to 0.53.0 and gazelle to 0.42.0.

With the explicit Go 1.23 SDK it should then be possible to update all the go deps including bringing `golang.org/x/crypto` up to v0.36.0. (And then rerun gazelle again).

andrii-korotkov-verkada:
@johnbartholomew, it worked! Thank you!

PR #799, opened 2025-04-06
Interpreter Stack Profiler flag

B13rg:
This is a continuation of #529 

I have attempted to finish what was started.

To not use global variable, I added the field `profilerOpts` to the VM and interpreter objects.
The `StackProfilerOpts` object stores what was the global variables.

As part of adding the field, I simplified the function calls of `evaluate`, `evaluateStream`, and `evaluateMulti` to just take the VM object instead of individual parameters.

I am unsure if it is the right type, or if it should be a `*StackProfilerOpts` instead.

Thanks!

### DIFF
--- a/interpreter.go
+++ b/interpreter.go
@@ -1339,13 +1339,14 @@ func buildObject(hide ast.ObjectFieldHide, fields map[string]value) *valueObject
 	return makeValueSimpleObject(bindingFrame{}, fieldMap, nil, nil)
 }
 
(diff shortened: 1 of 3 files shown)


interpreter.go
```diff
@@ -281,6 +286,15 @@ type interpreter struct {
 	stack callStack
 
 	evalHook EvalHook
+
+	profilerOpts StackProfilerOpts
```
B13rg, line 290: When running the `govet` linter, I am seeing the following issue:

```sh
interpreter.go:267:18: fieldalignment: struct with 128 pointer bytes could be 104 (govet)
type interpreter struct {
```

PR #842, opened 2026-01-30
fix linter to add optional parameters to std.manifestJsonEx

### LINKED ISSUE #583: std.manifestJsonEx - different arguments count than what documented
Hi, according to [official docs](https://jsonnet.org/ref/stdlib.html) there should be a `std.manifestJsonEx` function which accept 4 arguments
> std.manifestJsonEx(value, indent, newline, key_val_sep)

Sounds good.  Looking at the original implementation [it is the case](https://github.com/google/jsonnet/blob/997ecd93b37862e462cdeebef749008c8f2f7af8/stdlib/std.jsonnet#L1008):

>   manifestJsonEx(value, indent, newline='\n', key_val_sep=': ')::

But in go's implementation, the function accepts only [2 arguments](https://github.com/google/go-jsonnet/blob/12bd29d164b131a4cd84f22f1456fe37136abc6d/linter/internal/types/stdlib.go#L112)

> 		"manifestJsonEx":     g.newSimpleFuncType(stringType, "value", "indent"),

I believe, it should be aligned with the docs.

netomi:
This fixes #583 .

It add the additional optional parameters `newline` and `key_val_sep` to the linter std.

`value` and `indent` are required.

### DIFF
--- a/linter/internal/types/stdlib.go
+++ b/linter/internal/types/stdlib.go
@@ -122,7 +122,7 @@ func prepareStdlib(g *typeGraph) {
 		"manifestPython":       g.newSimpleFuncType(stringType, "v"),
 		"manifestPythonVars":   g.newSimpleFuncType(stringType, "conf"),
 		"manifestTomlEx":       g.newSimpleFuncType(stringType, "value", "indent"),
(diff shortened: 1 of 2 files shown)


He-Pin:
@netomi This pr missing a golden file for the result

PR #845, opened 2026-02-09
Chore: import golang golden overrides from cpp-jsonnet

johnbartholomew:
All the go-jsonnet golden output overrides should be kept here in the go-jsonnet repo. The older override mechanism (in which the cpp-jsonnet repo has golden files for Golang ending in .golang) will be removed.


CertainLach:
Jrsonnet is fine with either naming, right now both c++ and golang tests pass (with some exceptions and overrides)
https://github.com/CertainLach/jrsonnet/blob/master/tests/tests/cpp_test_suite.rs#L232-L242

The only thing I wish is that tests had some indications that they contain stacktrace in their name, similar to "error.*" tests in cpp-jsonnet, because stack traces are very different in jrsonnet/sjsonnet, and with some exceptions (golang's/cpp ugly float formatting) only they should be different between implementations

PR #855, opened 2026-02-28
chore: update stdlib_smoke_test.jsonnet with with float closing test

He-Pin:
This will help sjsonnet and jrsonnet

### DIFF
--- a/testdata/stdlib_smoke_test.jsonnet
+++ b/testdata/stdlib_smoke_test.jsonnet
@@ -4,6 +4,34 @@
 // Functions without optional arguments need only one line
 // Functions with optional arguments need two lines - one with none of the optional arguments
 // and the other with all of them.
+local assertClose(a, b) =
+  // Using 1e-12 as tolerance. Jsonnet uses double-precision floats with machine epsilon of 2**-53 ≈ 1.11e-16.
+  // This tolerance is ~9000x the machine epsilon, which is quite lenient and should work across
+  // different platforms and math libraries.
+  //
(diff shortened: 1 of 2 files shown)


testdata/stdlib_smoke_test.jsonnet
```diff
@@ -5,6 +5,17 @@
 // Functions with optional arguments need two lines - one with none of the optional arguments
 // and the other with all of them.
 
+local assertClose(a, b) =
+  local err =
+    if b == 0 then
+      a - b
```
johnbartholomew: If `b` is zero, then `a - b` is just `a`.
This also needs use use the absolute value; currently assertClose(-100, 0) == true

testdata/stdlib_smoke_test.jsonnet
```diff
@@ -5,6 +5,17 @@
...
+    else
+      if a / b - 1 > 0 then a / b - 1 else 1 - a / b;
+  if err > 0.000005 then
```
johnbartholomew: I would suggest a tighter bound here, probably set it to `1e-12` (nothing very special about this value, it's just what I would pick)

Jsonnet uses double-precision floats, the epsilon value for them is 2**-53 = 1.1102230246251565e-16. Even with platform/maths library differences, I would expect all implementations to be able to match down to relative error of 1e-12 or better; that's ~9000 times the machine epsilon which is quite lenient, if I've got my own maths right.

Having said that, if you're explicitly aiming for this to work for a (hypothetical) single-precision float implementation, then I could accept that. But if so, please add a comment to make the reason clear.

PR #869, opened 2026-04-03
feat: add --wrap flag to apply a jsonnet expression to the top-level value

### LINKED ISSUE #195: A commandline option for output format
We could provide a convenience option `--wrap` and it could be used as `jsonnet --wrap std.manifestYAML file.jsonnet` and it would pass whatever `file.jsonnet` evaluates to, into `std.manifestYAML` (and assume raw string output afterwards, I guess).

it would allow output in any format, including custom ones.

This way, for example `manifestINI`, which we already have, could be used in the same way
and even if someone decides they want to generate, say, nginx config, with their custom function, they could do it just as easily.

Is there a better name for this option than `--wrap`?

vivainio:
Closes #195.

## Motivation

My primary use case is emitting YAML output directly from jsonnet, without piping through an external converter. The `--wrap` flag makes this possible with `--wrap std.manifestYamlDoc`, while also being general enough to support any output format via stdlib or custom functions.

## Summary

Adds a `--wrap <expr>` flag that applies a jsonnet expression to the top-level evaluated value before output, with string output mode implied.

This is the "simple wrapping" variant discussed in #195 — it constructs a snippet `(expr)(value)` and evaluates it with string output. This is equivalent to the existing `jsonnet -S -e 'std.manifestINI(import "file.jsonnet")'` pattern but more convenient.

## Known limitation: stack traces

As discussed in #195, when an error occurs during manifestation through a stdlib function like `std.manifestYamlDoc`, the stack trace includes stdlib internals, which is noisier than errors during normal evaluation. A future enhancement could add a "deep evaluation" variant that fully evaluates the value before passing it to the manifester, producing cleaner error traces for the custom manifester use case.

## Examples

```sh
# YAML output
jsonnet --wrap std.manifestYamlDoc file.jsonnet

# YAML without quoted keys
jsonnet --wrap 'function(x) std.manifestYamlDoc(x, quote_keys=false)' file.jsonnet

# Pretty JSON via stdlib
jsonnet --wrap std.manifestJson file.jsonnet

# Extract a single field for use in shell scripts
jsonnet --wrap 'function(x) x.version' file.jsonnet
```

## Notes

- `--wrap` is incompatible with `--multi`, `--yaml-stream`, and `--string`
- For file input, builds `(expr)(import @'file')` as a snippet
- For `-e` / stdin, inlines the input as `(expr)(input)`
- The wrap expression must evaluate to a function of one argument

## Test plan

- [x] `jsonnet --wrap std.manifestYamlDoc file.jsonnet` produces YAML
- [x] `jsonnet --wrap 'function(x) std.manifestYamlDoc(x, quote_keys=false)' file.jsonnet` produces unquoted-key YAML
- [x] Works with stdin (`-`) and `-e` code input
- [x] `--wrap` + `--multi` / `--yaml-stream` / `--string` each produce a clear error

### DIFF
--- a/cmd/jsonnet/cmd.go
+++ b/cmd/jsonnet/cmd.go
@@ -52,6 +52,8 @@ func usage(o io.Writer) {
 	fmt.Fprintln(o, "                             files")
 	fmt.Fprintln(o, "  -y / --yaml-stream         Write output as a YAML stream of JSON documents")
 	fmt.Fprintln(o, "  -S / --string              Expect a string, manifest as plain text")
+	fmt.Fprintln(o, "  --wrap <expr>              Apply <expr> to the top-level value before output,")
+	fmt.Fprintln(o, "                             e.g. --wrap std.manifestYamlDoc")
 	fmt.Fprintln(o, "  --no-trailing-newline      Do not add a trailing newline to the output")
 	fmt.Fprintln(o, "  -s / --max-stack <n>       Number of allowed stack frames")
 	fmt.Fprintln(o, "  -t / --max-trace <n>       Max length of stack trace before cropping")
@@ -102,6 +104,7 @@ type config struct {
 	evalMulti            bool
 	evalStream           bool
 	evalCreateOutputDirs bool
+	wrapExpr             string
 }
 
 func makeConfig() config {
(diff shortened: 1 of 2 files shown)


He-Pin:
is there a pr in the cpp implementation too?

vivainio:
No, I wasn't aware this is needed

vivainio:
I am closing this for lack of personal need (using jrsonnet for this)

PR #876, opened 2026-05-22
Reject duplicate function parameters

### LINKED ISSUE #873: go-jsonnet does not reject duplicate parameter definition
```
(function(x, x, x) x)(null, 3, 2) // 2
```

This doesn't seem right

rohan-patnaik:
Rejects duplicate parameter names while parsing function and method parameter lists.

Previously, go-jsonnet accepted inputs like:

```jsonnet
(function(x, x, x) x)(null, 3, 2)
```

That makes argument/default handling ambiguous and differs from other Jsonnet implementations, which reject duplicate parameters.

This change tracks parameter names while parsing and returns a static error when the same name appears twice. It covers normal functions, local function sugar, and object methods.

Fixes #873.

Testing:
- `go test ./internal/parser`
- `go test ./...`

### DIFF
--- a/internal/parser/parser.go
+++ b/internal/parser/parser.go
@@ -198,6 +198,7 @@ func (p *parser) parseParameters(elementKind string) (*token, []ast.Parameter, b
 
 	var parenR *token
 	var params []ast.Parameter
+	seenParams := ast.NewIdentifierSet()
 	gotComma := false
 	first := true
 	for {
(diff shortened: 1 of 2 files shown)


He-Pin:
Great， does the same problem exist in cpp?

CertainLach:
CPP is fine

rohan-patnaik:
@johnbartholomew, could you take a look at this duplicate-parameter fix when convenient? Please let me know if anything else is needed from me. Thanks!

PR #877, opened 2026-05-26
Document JSONNET_PATH default behavior

### LINKED ISSUE #659: Reference to what the default JSONNET_PATH is
I can't find any information about what the default JSONNET_PATH is. This is a request to update the documentation

Haihan-Jiang:
## What

Adds a short README section for import search paths. It makes the default behavior explicit: the command line tools do not add library search directories unless the user provides `JSONNET_PATH` or `-J` / `--jpath`, and imports are resolved relative to the importing file first.

The section also mirrors the existing CLI help for path separators and precedence so users can find it without running `--help`.

Fixes #659.

## Testing

- `git diff --check -- README.md`
- `go test ./...`
- checked the documented `JSONNET_PATH` / `-J` precedence with a temporary local import example

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,25 @@ It's also available on Homebrew:
 brew install go-jsonnet
 ```
 
+## Import search path
+
+By default, the command line tools do not add any library search directories.
+Imports are first resolved relative to the file that contains the import. Extra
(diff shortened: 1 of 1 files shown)


Haihan-Jiang:
This is ready for review from my side. Could someone take a look when convenient?

PR #884, opened 2026-06-26
fix: std.format accepts boolean for numeric conversion codes

He-Pin:
## Summary

The [official Jsonnet spec](https://jsonnet.org/ref/stdlib.html#std-format) states:
> The string formatting follows the same rules as Python.

In Python, `bool` is a subclass of `int`, so `"%d" % True` returns `"1"`. However, go-jsonnet currently rejects booleans for all numeric conversion codes with `Format required number at N, got boolean`.

This fix coerces boolean values to `0`/`1` in `format_code()` before numeric conversion, matching Python's behavior.

Upstream PR: google/jsonnet#1328

### Before / After

| Expression | Python | go-jsonnet (before) | go-jsonnet (after) |
|---|---|---|---|
| `"%d" % true` | `"1"` | ERROR | `"1"` |
| `"%f" % false` | `"0.000000"` | ERROR | `"0.000000"` |
| `"%x" % true` | `"1"` | ERROR | `"1"` |
| `"%s" % true` | `"True"` | `"true"` | `"true"` |

### Changes

- `stdlib/std.jsonnet`: Added boolean-to-number coercion in `format_code()`. The `%s` branch is handled first (before coercion) so `%s` with `true` still returns `"true"`. All numeric branches (`d`, `o`, `x`, `f`, `e`, `g`, `c`) use `num_val` which coerces booleans to `0`/`1`.
- `astgen/stdast.go`: Regenerated from the updated `stdlib/std.jsonnet`.
- Golden files: Updated to reflect shifted line numbers in error traces.

## Test plan
- [x] Existing format tests continue to pass (`go test ./...` passes)
- [x] Verified `%d`, `%f`, `%x`, `%o`, `%e`, `%g` with boolean values produce correct output
- [x] Verified `%s` with boolean values still returns string representation


CertainLach:
Given that this change is dubious from correctness standpoint as explained in https://github.com/deltarocks/jrsonnet/pull/233, was rejected in jrsonnet, being rolled-forward in sjsonnet (https://github.com/databricks/sjsonnet/pull/1069), and is not implemented in other impls, this PR should be closed

PR #892, opened 2026-07-23
fix: add stack depth accounting to rawEquals

prasanna8585:
## Problem

`rawEquals` in `builtins.go`, which implements jsonnet's `==` and `!=` operators, recurses natively via the Go call stack when comparing nested arrays or objects. It has no stack-depth accounting anywhere in the call chain.

A self-referential jsonnet value compared with `==`:
```
{a: $} == {a: $}
```
causes `rawEquals` to recurse into itself indefinitely, consuming memory until the process is killed by the OS (OOM) or the Go runtime's stack limit is hit.

## Precedent

This is the same bug class already fixed in `builtinManifestJSONEx`, `builtinManifestYamlDoc`, and `builtinManifestTomlEx` (depth-counter pattern seeded from `i.stack.limit`, decremented per recursive call, erroring with "max manifest depth exceeded"). `rawEquals` was never covered by that sweep and remains vulnerable to the identical trigger pattern.

## Fix

Adds a `depth int` parameter to `rawEquals`, seeded from `i.stack.limit` (the interpreter's own configured stack limit, default 500) at all 5 entry points, and decremented on each of the 2 recursive call sites (array element comparison, object field comparison). Errors with `"max equality depth exceeded, possible infinite recursion"` when exhausted — matching the existing manifest-builtin error message style exactly.

## Testing

Verified with `{a: $} == {a: $}` — with the fix, this now returns a clean error instead of hanging indefinitely with unbounded memory growth.

### DIFF
--- a/builtins.go
+++ b/builtins.go
@@ -614,7 +614,7 @@ func rawMember(i *interpreter, arrv, value value) (bool, error) {
 			if err != nil {
 				return false, err
 			}
-			equal, err := rawEquals(i, cachedThunkValue, value)
+			equal, err := rawEquals(i, cachedThunkValue, value, i.stack.limit)
 			if err != nil {
 				return false, err
 			}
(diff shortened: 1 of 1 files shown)


prasanna8585:
Hi @johnbartholomew @rohitjangid,

Just checking in to see if you've had a chance to look at the patch strategy for this. I'm happy to spin up the code changes for the fix in a private fork if that helps speed things up.

Thanks again for your time!

prasanna8585:
Hi @johnbartholomew @rohitjangid,

Just checking in to see if you've had a chance to look at the patch strategy for this. I'm happy to spin up the code changes for the fix in a private fork if that helps speed things up.

Thanks again for your time!

PR #898, opened 2026-09-18
fix(parser): accept CRLF line endings in ||| text blocks

### LINKED ISSUE #786: go-jsonnet windows version is not compatible with grafonnet
Until v0.20.0, go-jsonnet windows x86-64 does not support syntax like:

This makes it imcompatible with grafonnet.

It seems this problem has been raised long time ago, but it seems still a problem on windows.
[Allow empty lines in multi-line blocks](https://github.com/google/jsonnet/issues/82)

vulragrag-star:
## Summary
Text blocks (`|||…|||`) failed to parse when the input used CRLF line endings: a blank line inside the block left a leading `\r` that broke the indent check and surfaced as `Text block not terminated with |||`. That matches the Windows / grafonnet report in #786 (also reproducible on Linux with CRLF input).

Treat `\r\n` as a single line terminator in the text-block lexer path and normalize recorded newlines to `\n`, so LF and CRLF inputs produce the same string value.

## Test plan
- [x] `go test ./internal/parser/ -run TextBlock`
- [x] `go test ./internal/parser/` and `go test .`
- [x] Manual: CRLF snippet from #786 now evaluates to `{"test": "Hello\n\nWorld\n"}` (same as LF)

Fixes #786


vulragrag-star:
I signed the CLA.

